PR #4, opened 2022-09-13
Refactor data api & datastore 

aratz-lasa:
# What 🕵️‍♀️
Simplifies the implementation of the DataAPI and Datastore
+ it implements a fix for #16 

# Why 🔑
This is important for delevopment speed and readability

# How ⚙️
I merged all the different data API calls into a single call, by creating a new parameter which contains all queryable parameters. 
For the datastore I did the same for the getHeader calls.

# Testing 🧪
I updated the corresponding test files


pkg/datastore.go
```diff
@@ -185,6 +175,28 @@ func (s DefaultDatastore) GetRegistration(ctx context.Context, pk PubKey) (types
 	return registration, err
 }
 
+func (s DefaultDatastore) queryToKey(ctx context.Context, query HeaderQuery) (ds.Key, error) {
+	var (
+		rawKey []byte
+		err    error
+	)
+
+	if strings.Compare(query.BlockHash.String(), "0x0000000000000000000000000000000000000000000000000000000000000000") != 0 {
```
lthibault, line 184: This should go in some kid of `Validate()` method (or similar) in the `HeaderQuery` type.  Same with the other checks.

lthibault (changes requested):
Looking cleaner and cleaner.  🙂

pkg/service.go
```diff
@@ -372,6 +372,8 @@ func (s *DefaultService) GetPayloadDelivered(ctx context.Context, query TraceQue
 
 	if err == nil {
 		return []types.BidTrace{event.BidTrace}, err
+	} else if errors.Is(err, ds.ErrNotFound) {
+		return []types.BidTrace{}, nil
```
lthibault, line 376: Can we return `nil` instead of allocating a `[]types.BidTrace{}`?

pkg/service.go
```diff
@@ -436,6 +438,8 @@ func (s *DefaultService) GetBlockReceived(ctx context.Context, query TraceQuery)
 
 	if err == nil {
 		return []BidTraceWithTimestamp{*event.Trace}, err
+	} else if errors.Is(err, ds.ErrNotFound) {
+		return []BidTraceWithTimestamp{}, nil
```
lthibault, line 442: Return `nil` instead of allocating?

lthibault (changes requested):

pkg/api.go
```diff
@@ -234,34 +234,22 @@ func (a *API) specificRegistration(w http.ResponseWriter, r *http.Request) (int,
 
 func (a *API) proposerPayloadsDelivered(w http.ResponseWriter, r *http.Request) (int, error) {
 	slot, err := specificSlot(r)
-	if err == nil {
-		blocks, err := a.Service.GetDelivered(r.Context(), slot)
-		return a.respond(w, blocks, err)
-	} else if err != nil && !errors.Is(err, ErrParamNotFound) {
+	if err != nil && !errors.Is(err, ErrParamNotFound) {
```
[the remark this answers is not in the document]
aratz-lasa: Yeah, so basically, checks whether the parameter exists. If not it raises the `ErrParamNotFound` and returns the zero value for the parameter. For example, it may not contain the slot parameter, so it will return the error and `Slot(0)`. We could use a boolean instead, but then, we would not catch other errors like parsing.

pkg/service.go
aratz-lasa: Not really, since we must return an empty slice to comply with other relays

Issue #6, opened 2022-09-14
getPayload has insufficient checks, allows validator to access the block contents without slashing


metachris:
The getPayload checks in https://github.com/blocknative/dreamboat/blob/main/pkg/api.go#L172 / https://github.com/blocknative/dreamboat/blob/main/pkg/relay.go#L312 are insufficient, and allow a validator to steal the payload without risk of slashing. besides checking the blockhash, also the correct slot and proposer index need to be verified.

See the mev-boost audit for more details: https://github.com/flashbots/mev-boost/blob/main/docs/audit-20220620.md#relayer-signedblindedbeaconblock-validation

> Relayer SignedBlindedBeaconBlock validation
>
> To be slashed, a validator would need to sign two competing blocks for the block of the slot they are in charge of producing.
>
> As such, relayers need to validate the SignedBlindedBeaconBlock signed by validators to be of the correct Slot and ProposerIndex before showing the unblinded block to the validator. If this is not checked, validators will be able to access the block contents and use these transactions to craft a new block, where they extract the MEV for themselves.

aratz-lasa:
Thank you very much for opening this issue @metachris . I just opened a PR to fix this issue!

lthibault:
Fixed by #7 

@metachris Thanks for the assist -- much appreciated! 🤝

PR #7, opened 2022-09-14
improved getPayload security

aratz-lasa:
# What 🕵️‍♀️
This PR improves getPayload security based on the issue #6 

# Why 🔑
In order to avoid a validator to steal the payload without risk of slashing

# How ⚙️
Checking the slot and proposer pubkey, apart from the BlockHash when retriving the payload from the datastore

# Testing 🧪
All the tests have been updated based on this new security improvement


lthibault (approved):

dmarzzz (approved):
Not a fan of `PayloadKeyKey` but as this is an artifact of maintaining our own index for the key-value store LGTM!

PR #9, opened 2022-09-14
fix cursor parameter

aratz-lasa:
# What 🕵️‍♀️
This PR fixes the `cursor` parameter behavior on the Data API

# Why 🔑
This is necessary for correct live auditing of the relay


lthibault (approved):
LGTM. Send it.

dmarzzz (approved):
LGTM

PR #11, opened 2022-09-14
Fix/get delivered payload

aratz-lasa:
# What 🕵️‍♀️
This PR fixes the Data API by using the correct block hash. Currently we are using the Blockhash from BidTrace of the block submission, but this is not correct. We should be using the ExecutionPayload Blockhash


dmarzzz (approved):
Good catch

PR #12, opened 2022-09-14
fixed delivered payload Data API

aratz-lasa:
# What 🕵️‍♀️
This PR fixes the payload delivery Data API. Previously the delivered payload was overwritten whenever a new block with the same slot was submitted. But now they are not. Delivered payloads and submitted blocks are stored separately.


dmarzzz (approved):
LGTM

PR #13, opened 2022-09-15
Fix/blocks received

aratz-lasa:
# What 🕵️‍♀️
This PR fixes the builder_blocks_received Data API call, so that now all blocks for every slot are returned. previously just the latest block per slot was returned

# Testing 🧪
The corresponding tests have been fixed, and all pass

### DIFF
--- a/pkg/api.go
+++ b/pkg/api.go
@@ -285,9 +285,16 @@ func (a *API) proposerPayloadsDelivered(w http.ResponseWriter, r *http.Request) 
 }
 
 func (a *API) builderBlocksReceived(w http.ResponseWriter, r *http.Request) (int, error) {
+	limit, err := limit(r)
+	if err != nil && !errors.Is(err, ErrParamNotFound) {
+		return http.StatusBadRequest, err
+	} else if errors.Is(err, ErrParamNotFound) {
+		limit = 100
+	}
+
 	slot, err := specificSlot(r)
 	if err == nil {
-		blocks, err := a.Service.GetBlockReceived(r.Context(), slot)
+		blocks, err := a.Service.GetBlockReceived(r.Context(), slot, limit)
 		return a.respond(w, blocks, err)
 	} else if err != nil && !errors.Is(err, ErrParamNotFound) {
 		return http.StatusBadRequest, err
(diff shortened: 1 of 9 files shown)


pkg/api.go
```diff
@@ -285,9 +285,16 @@ func (a *API) proposerPayloadsDelivered(w http.ResponseWriter, r *http.Request)
 }
 
 func (a *API) builderBlocksReceived(w http.ResponseWriter, r *http.Request) (int, error) {
+	limit, err := limit(r)
+	if err != nil && !errors.Is(err, ErrParamNotFound) {
+		return http.StatusBadRequest, err
+	} else if errors.Is(err, ErrParamNotFound) {
```
lthibault, line 291: We can remove the `else`, since we're returning above.

pkg/datastore.go
```diff
@@ -105,9 +116,18 @@ func (s DefaultDatastore) GetHeaderByBlockHash(ctx context.Context, bh types.Has
...
-	var trace HeaderAndTrace
-	err = json.Unmarshal(data, &trace)
-	return trace, err
+	var headers []HeaderAndTrace
+	if err = json.Unmarshal(data, &headers); err != nil {
+		return HeaderAndTrace{}, err
+	}
+
+	for _, header := range headers {
+		if strings.Compare(header.Header.BlockHash.String(), bh.String()) == 0 {
```
lthibault, line 125: This can be simplified to `if header.Header.BlockHash == bh`.  We'll get a small performance win, too.

lthibault (changes requested):

pkg/datastore.go
dmarzzz: what about equalFold 👀👀 https://dzone.com/articles/optimizing-string-comparisons-in-go

dmarzzz (approved):
other than above changes LGTM!

pkg/datastore.go
lthibault: I don't think we care so much about the performance win (my fault for bringing it up 😆) as the legibility. 

Also, I saw a bunch of similar string comparisons littered around the codebase.   We should definitely fix those as well.

aratz-lasa: Thank you for the review gents. But many of these comparisons are optimized in the refactor data API PR. So after merging that this will be fixed

dmarzzz: LGTM then!

dmarzzz:
Closing this PR as we're planning a larger data API refactor in the near future

aratz-lasa:
Opened #21 instead

PR #14, opened 2022-09-15
fix registration logic

aratz-lasa:
# What 🕵️‍♀️
This PR improves validator registration logic. Currently, as soon as registration fails, the remaining validators are not registered. However, with this PR even if a single registration fails, then the rest of validators are registered.

### DIFF
--- a/pkg/config.go
+++ b/pkg/config.go
@@ -36,7 +36,7 @@ type PubKey struct{ types.PublicKey }
 
 func (pk PubKey) Loggable() map[string]any {
 	return map[string]any{
-		"pub_key": pk,
+		"pubkey": pk,
 	}
 }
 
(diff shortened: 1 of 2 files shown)


metachris:
We've also considered it, but this change has two major problems:
- First, it doesn't provide the proposer any feedback which proposer failed to register (but they should know)
- Second, this opens you up to DoS attacks, where a single request can include a large number of forged registrations with invalid signature, and they will all need to get verified, taking about 1ms CPU time per request. 10k requests will take 10 CPUs 1 second to process. I guess you can see how this can be abused.

dmarzzz:
Thanks for the input @metachris , greatly appreciated! 🙏🙏 Closing this PR @aratz-lasa as it is undesired behaviour and improving the return msg for failed registrations is included in #15

PR #15, opened 2022-09-16
Stable/log+bid

aratz-lasa:
# What 🕵️‍♀️
This PR:

- Cleans up logs, since previously it was too noisy
- Does not accept blocks after payload is delivered. This allows to signal the builder to stop building blocks for that slot


pkg/relay.go
```diff
@@ -405,15 +408,18 @@ func SubmitBlockRequestToSignedBuilderBid(req *types.BuilderSubmitBlockRequest,
...
 
-	logger.With(log.F{
-		"processingTimeMs": time.Since(timeStart).Milliseconds(),
-		"slot":             submitBlockRequest.Message.Slot,
-		"blockHash":        submitBlockRequest.ExecutionPayload.BlockHash,
-		"proposer":         submitBlockRequest.Message.ProposerPubkey,
-	}).Trace("block submission requested")
+	logger := rs.Log().With(log.F{
+		"method":    "SubmitBlock",
+		"builder":   submitBlockRequest.Message.BuilderPubkey,
```
lthibault, line 415: This is big enough that I think a `Loggable` wrapper type for `types.BuilderSubmitBlockRequest` is warranted.

pkg/relay.go
```diff
@@ -440,6 +446,18 @@ func (rs *DefaultRelay) SubmitBlock(ctx context.Context, submitBlockRequest *typ
 
 		return fmt.Errorf("block submission failed: %w", err)
 	}
+
+	slot := Slot(submitBlockRequest.Message.Slot)
```
lthibault, line 450: While we're at it, we should add a `.Slot()` method to the above wrapper type.

lthibault (changes requested):
Requesting two small code-quality changes, but otherwise 👍

dmarzzz:
Will review all these shortly but Im inclined to revert the increasing bid per slot constraint in favor of this comment https://github.com/flashbots/mev-boost/issues/113#issuecomment-1209944942

pkg/relay.go
```diff
@@ -123,7 +124,7 @@ func (rs *DefaultRelay) RegisterValidator(ctx context.Context, payload []types.S
 }
 
 func (rs *DefaultRelay) processValidator(ctx context.Context, payload []types.SignedValidatorRegistration, state State) error {
-	logger := rs.Log().WithField("method", "RegisterValidator")
+	logger := rs.Log().WithField("method", "ProcessValidator")
```
dmarzzz, line 127: I dont think we need to rename this. Having it under registerValdiator lets us associate it with other logs that come from the higher up function. log body messages should be sufficient to delineate

pkg/relay.go
```diff
@@ -135,12 +136,15 @@ func (rs *DefaultRelay) processValidator(ctx context.Context, payload []types.Si
 			registerRequest.Signature[:],
 		)
 		if !ok || err != nil {
-			logger.WithError(err).Debug("signature invalid")
-			return fmt.Errorf("signature invalid")
+			logger.
+				WithError(err).
+				WithField("pubkey", registerRequest.Message.Pubkey).
+				Debug("signature invalid")
+			return fmt.Errorf("signature invalid for %s", registerRequest.Message.Pubkey.String())
```
dmarzzz, line 143: great add and solves issue brought up in #14

pkg/relay.go
```diff
@@ -150,50 +154,52 @@ func (rs *DefaultRelay) processValidator(ctx context.Context, payload []types.Si
 			return err
 		} else if !ok {
 			if rs.config.CheckKnownValidator {
-				return fmt.Errorf("not a validator")
+				return fmt.Errorf("%s not a known validator", registerRequest.Message.Pubkey.String())
 			} else {
-				logger.WithField("pubkey", pk.PublicKey).Debug("is not a known validator")
+				logger.
+					WithField("pubkey", pk.PublicKey).
```
dmarzzz, line 160: pubkey -> pub_key consistent w/ above changes

pkg/relay.go
```diff
@@ -150,50 +154,52 @@ func (rs *DefaultRelay) processValidator(ctx context.Context, payload []types.Si
...
 				(registerRequest.Message.FeeRecipient != previousValidator.Message.FeeRecipient ||
 					registerRequest.Message.GasLimit != previousValidator.Message.GasLimit) {
 				// to help debug issues with validator set ups
-				rs.Log().With(log.F{
+				logger.With(log.F{
 					"prevFeeRecipient":    previousValidator.Message.FeeRecipient,
 					"requestFeeRecipient": registerRequest.Message.FeeRecipient,
 					"prevGasLimit":        previousValidator.Message.GasLimit,
 					"requestGasLimit":     registerRequest.Message.GasLimit,
+					"pubkey":              registerRequest.Message.Pubkey,
```
dmarzzz, line 188: pubkey -> pub_key

pkg/relay.go
```diff
@@ -150,50 +154,52 @@ func (rs *DefaultRelay) processValidator(ctx context.Context, payload []types.Si
...
 		}
-
-		logger.WithField("pubkey", pk).Trace("validator registered")
 	}
 
 	logger.With(log.F{
-		"processingTimeMs": time.Since(timeStart).Milliseconds(),
-		"numberValidators": len(payload),
+		"processingTimeMs":      time.Since(timeStart).Milliseconds(),
+		"numberBatchValidators": len(payload),
```
dmarzzz, line 202: I don't think this is necessary for the same reason as above, mainly we can discern this is for a batch by the message

pkg/relay.go
```diff
@@ -306,8 +314,8 @@ func (rs *DefaultRelay) GetPayload(ctx context.Context, payloadRequest *types.Si
 		payloadRequest.Signature[:],
 	)
 	if !ok || err != nil {
-		rs.Log().WithField(
-			"pubKey", proposerPubkey,
+		logger.WithField(
+			"proposer", proposerPubkey,
```
dmarzzz, line 318: can we make this pub_key and the msg say "proposer signature invalid. We should actually value that convention across the rest of the repo

dmarzzz (changes requested):

pkg/relay.go
aratz-lasa: But in all the places we are using `pubkey` instead of `pub_key`. Also, this way we are consistent with JSON fields that are written `pubkey` instead of `pub_key`

dmarzzz: sorry i misread the above change

pkg/relay.go
```diff
@@ -440,6 +454,24 @@ func (rs *DefaultRelay) SubmitBlock(ctx context.Context, submitBlockRequest *typ
...
+
+	slot := Slot(submitBlockRequest.Message.Slot)
+
+	_, err = state.Datastore().GetDelivered(ctx, slot)
+	if err == nil {
+		logger.Debug("block submission after payload delivered")
+		return errors.New("the slot payload was already delivered")
+	}
+
+	existingHeader, err := state.Datastore().GetHeader(ctx, slot)
```
dmarzzz, line 466: can we remove this?

pkg/service.go
```diff
@@ -267,35 +268,28 @@ func (s *DefaultService) updateProposerDuties(ctx context.Context, client Beacon
...
 	s.state.duties.Store(state)
 
 	logger.With(log.F{
-		"epochFrom":        epoch,
-		"epochTo":          epoch + 1,
-		"processingTimeMs": time.Since(timeStart).Milliseconds(),
-		"receivedDuties":   len(current.Data),
+		"processingTimeMs":        time.Since(timeStart).Milliseconds(),
+		"receivedDuties":          len(entries),
+		"notRegisteredValidators": notRegistered,
```
dmarzzz, line 292: what is the purpose of not registered validators? I think we need the opposite and log the registered validators number

aratz-lasa: You are right. Moreover, we can calculate the amount of registered validators based on the length of `state.proposerDutiesResponse. `. FIxed.

PR #24, opened 2022-10-11
Fix/improve Data API builder_blocks_received

aratz-lasa:
# What 🕵️‍♀️
This PR 
(1) deduplicates Headers. Currently, headers are stored as an array, and when we add a new one we simply append to the list. However, this may lead to having hundreds of duplicate headers, which obscures the Data API. So, in this PR we add the logic for deduplicating the headers.

(2) Fixes the data API. Currently we are not correctly serving the indexed queries. We return more than what user asks for.

# How ⚙️
Before appending a new header we check if the tail of the headers is the same as the one currently being stored. For checking equality only the `BlockHash` is used.

# Testing 🧪
I added a regression test for checking deduplication works as expected

### DIFF
--- a/pkg/datastore.go
+++ b/pkg/datastore.go
@@ -127,7 +127,7 @@ func (s *DefaultDatastore) getHeaders(ctx context.Context, key ds.Key) ([]Header
 }
 
 func (s *DefaultDatastore) filterHeaders(headers []HeaderAndTrace, query Query) []HeaderAndTrace {
-	var filtered []HeaderAndTrace
+	filtered := headers[:0]
 	for _, header := range headers {
 		if (query.BlockHash != types.Hash{}) && (query.BlockHash != header.Header.BlockHash) {
 			continue
--- a/pkg/datastore.go
+++ b/pkg/datastore.go
@@ -83,6 +83,10 @@ func (s *DefaultDatastore) PutHeader(ctx context.Context, slot Slot, header Head
 		return err
 	}
 
+	if 0 < len(headers) && headers[len(headers)-1].Header.BlockHash == header.Header.BlockHash {
+		return nil // deduplicate
+	}
+
 	headers = append(headers, header)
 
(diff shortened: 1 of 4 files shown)


pkg/datastore.go
```diff
@@ -117,6 +126,27 @@ func (s *DefaultDatastore) getHeaders(ctx context.Context, key ds.Key) ([]Header
 	return s.unsmarshalHeaders(data)
 }
 
+func (s *DefaultDatastore) filterHeaders(headers []HeaderAndTrace, query Query) []HeaderAndTrace {
+	var filtered []HeaderAndTrace
+	for _, header := range headers {
```
lthibault, line 131: This would be a good place to do some [zero-allocation filtering](https://zchee.github.io/golang-wiki/SliceTricks/#filtering-without-allocating)

pkg/datastore.go
```diff
@@ -185,30 +215,25 @@ func (s *DefaultDatastore) GetDeliveredBatch(ctx context.Context, queries []Quer
 }
 
 func (s *DefaultDatastore) GetHeaderBatch(ctx context.Context, queries []Query) ([]HeaderAndTrace, error) {
-	keys := make([]ds.Key, 0, len(queries))
+	var batch []HeaderAndTrace
```
lthibault, line 218: Same as above.  Let's avoid that alloc and keep this thing faaaaaaaast.

lthibault (changes requested):

pkg/datastore.go
aratz-lasa: Done 👍

pkg/datastore.go
aratz-lasa: How can we do that in this case exactly?

lthibault: I misread.  Please disregard

lthibault (approved):

dmarzzz (approved):
LGTM - one small note, I personally think 'filtering duplicates' === 'deduplicate'

lthibault (approved):

PR #29, opened 2022-10-21
Add correct slot ids

lukanus:
# What 🕵️‍♀️
Fixes broken tests, that was assigning 2 different values and then testing it's equality

# Why 🔑
Broken tests


dmarzzz (approved):
LGTM ty!

PR #30, opened 2022-10-21
Add num_tx and block_number to payload

lukanus:
# What 🕵️‍♀️
Adds num_tx and block_number to payload

# Why 🔑
Consistency after (https://collective.flashbots.net/t/relays-data-api-include-block-number-and-num-tx/527/2)


dmarzzz (approved):
LGTM!

PR #31, opened 2022-10-24
Add extensibility for adding arbitrary networks

aratz-lasa:
# What 🕵️‍♀️
This PR adds extensibility, by letting users to specify any arbitrary networks

# How ⚙️
For adding new networks, users MUST do 1 things:

- Write the network configuration in JSON format under `<datadir>/networks.json`

A simple example of a network specification file would be:
```
{
    "blocknative":{
        "GenesisForkVersion": "",
        "GenesisValidatorsRoot":"",
        "BellatrixForkVersion": ""
    },
}
```
This is only in case you want to extend the networks. The default networks (e.g., mainnet) are hardcoded.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,9 +55,40 @@ The following features and enhancements are in-progress.
 - [x] Runtime profiling & tracing endpoint
 - [x] Parallel validator-signature verification
 - [x] Contention-free execution along hot paths
-- [ ] Support for multiple beacon clients (in testing stage)
+- [x] Support for multiple beacon clients
 - [ ] Support for multiple external block builders
 
+## Extend the blockchain networks
+Do you need to make your relay work with a different blockchain that is not:
+- Mainnet
+- Kiln
+- Ropsten
+- Sepolia
+- Goerli
+
+Then you need to extend the relay by creating a json file called `networks.json`, and put it inside `datadir` (`datadir` is a parameter you set when running the relay).
+
+Inside the network.json file, you need to specify the network name and three other information:
+- GenesisForkVersion
+- GenesisValidatorsRoot
+- BellatrixForkVersion
+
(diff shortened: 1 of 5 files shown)


dmarzzz:
Why are we doing this through a JSON? That will increase operations complexity by requiring infrastructure tools to provision JSON files before running the application

lthibault (approved):

aratz-lasa:
@dmarzzz In my opinion, and after talking to Ayaz, the JSON simplifies, since the user does not need to know mainnet and other default network forkVersion, bellatrixVersion, etc. Since the most popular networks are harcoded in code. Moreover, they can add multiple networks at once in a single JSON.

dmarzzz:
what does adding multiple networks in one JSON mean here?

aratz-lasa:
> what does adding multiple networks in one JSON mean here?

It means you can extend the relay with multiple network specifications with a single JSON. So you could do:
```
{
    "blocknative":{
        "GenesisForkVersion": "",
        "GenesisValidatorsRoot":"",
        "BellatrixForkVersion": ""
    },
    "bitcoin":{
        "GenesisForkVersion": "",
        "GenesisValidatorsRoot":"",
        "BellatrixForkVersion": ""
    },
}
```
 And later use the network you prefer, by simply setting the `network` parameter by its name.

dmarzzz:
1. Can you add error checking that the values are at least hex
2. Can you add documentation to the ReadMe explaining how to use this

aratz-lasa:
@dmarzzz Done. Let me know, if I should explain something else in the `README`

dmarzzz:
Perfect thank you @aratz-lasa 🙏🙏 For posterity, we discussed in DM that we don't need to check that network values are hex since that will bubble up when computing the domain.

PR #33, opened 2022-11-03
Additional metric set

lukanus:


pkg/metrics/datastore.go
```diff
@@ -0,0 +1,25 @@
+package metrics
+
+import (
+	met "github.com/blocknative/dreamboat/metrics"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+func InitBadgerMetrics(m *met.Metrics) error {
+	a := map[string]*prometheus.Desc{
+		"badger_v2_blocked_puts_total":   prometheus.NewDesc("badger_blocked_puts_total", "Blocked Puts", nil, nil),
```
dmarzzz, line 10: can we rename these to datastore instead of badger so that its future proof for when we move to a generic data store implementation

dmarzzz:
also description plus testing plz 😄

cmd/dreamboat/main.go
```diff
@@ -213,6 +221,30 @@ func run() cli.ActionFunc {
 			return err
 		})
 
+		// run the http server
```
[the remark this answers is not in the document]
lukanus: Yeah I'll change

pkg/metrics/datastore.go
lukanus: Renamed

lukanus:
This one would be included in the reorg, so closing this separate one

PR #34, opened 2022-11-04
Wait until beacon node is synced on initialization

aratz-lasa:
# What 🕵️‍♀️
This PR changes relay's initialisation behaviour. If the beacon node is syncing, instead of failing, it loops until the beacon node is synced. It sleeps 3 seconds before querying again the beacon. I think there is no need for using a more sophisticated backoff mechanism.

### DIFF
--- a/pkg/service.go
+++ b/pkg/service.go
@@ -96,12 +96,9 @@ func (s *Service) setReady() {
 func (s *Service) RunBeacon(ctx context.Context, client BeaconClient) error {
 	logger := s.Log.WithField("method", "RunBeacon")
 
-	syncStatus, err := client.SyncStatus()
+	syncStatus, err := s.waitSynced(ctx, client)
 	if err != nil {
 		return err
-	}
-	if syncStatus.IsSyncing {
-		return ErrBeaconNodeSyncing
 	}
 
 	genesis, err := client.Genesis()
@@ -149,6 +146,24 @@ func (s *Service) RunBeacon(ctx context.Context, client BeaconClient) error {
 				"numKnownValidators":        len(validators),
 				"knownValidatorsUpdateTime": s.knownValidatorsUpdateTime(),
 			}).Debug("processed new slot")
+		}
+	}
+}
+
+func (s *Service) waitSynced(ctx context.Context, client BeaconClient) (*SyncStatusPayloadData, error) {
+	logger := s.Log.WithField("method", "WaitSynced")
+
+	for {
+		status, err := client.SyncStatus()
+		if err != nil || !status.IsSyncing {
+			return status, err
+		}
+
+		logger.Debug("beacon clients are syncing...")
+		select {
+		case <-ctx.Done():
+			return nil, ctx.Err()
+		case <-time.After(3 * time.Second):
 		}
 	}
 }


pkg/service.go
```diff
@@ -165,17 +165,23 @@ func (s *DefaultService) setReady() {
...
-	}
-	if syncStatus.IsSyncing {
-		return ErrBeaconNodeSyncing
-	}
+	for {
+		syncStatus, err := client.SyncStatus()
+		if err != nil {
+			return err
+		}
+		if syncStatus.IsSyncing {
```
lthibault, line 173: How difficult would it be to define an async `Synchronize(ctx context.Context) error` method on `BeaconClient`?  It would be much cleaner.

aratz-lasa: I don't think the complexity is worth it in this case

aratz-lasa: Okay, done

pkg/beacon.go
```diff
@@ -76,6 +77,22 @@ func (b *MultiBeaconClient) GetProposerDuties(epoch Epoch) (*RegisteredProposers
...
 }
 
+func (b *MultiBeaconClient) WaitSynced(ctx context.Context) (*SyncStatusPayloadData, error) {
+	for {
+		status, err := b.SyncStatus()
+		if err != nil || !status.IsSyncing {
+			return status, err
+		}
+
+		b.Log.Debug("bacon clients are syncing...")
```
lthibault, line 87: Mmmmm ... bacon ... 🥓 😋

pkg/beacon.go
```diff
@@ -231,6 +248,24 @@ func (b *beaconClient) GetProposerDuties(epoch Epoch) (*RegisteredProposersRespo
...
 
+// WaitSynced waits until beacon is synced of context is cancelled
+func (b *beaconClient) WaitSynced(ctx context.Context) (*SyncStatusPayloadData, error) {
+	for {
+		status, err := b.SyncStatus()
+		if err != nil || !status.IsSyncing {
+			return status, err
+		}
+
+		b.log.Debug("bacon client is syncing...")
```
lthibault, line 259: 🥓

lthibault (changes requested):
Minor typo: see below.

pkg/service.go
```diff
@@ -153,6 +150,24 @@ func (s *Service) RunBeacon(ctx context.Context, client BeaconClient) error {
...
+func (s *Service) waitSynced(ctx context.Context, client BeaconClient) (*SyncStatusPayloadData, error) {
+	logger := s.Log.WithField("method", "WaitSynced")
+
+	for {
+		status, err := client.SyncStatus()
+		if err != nil || !status.IsSyncing {
+			return status, err
+		}
+
+		logger.Debug("beacon clients are syncing...")
```
lukanus, line 162: Do we need that logged every minute? asking for a friend ;>

lukanus (approved):

pkg/service.go
aratz-lasa: This is only logged in the startup, until it's first synced. This way the user knows on startup when the beacon is synced and that the relay can start accepting requests

lukanus: a yea you're right, ok then

PR #35, opened 2022-11-08
propFeeRecipient from bidTrace not exec payload

dmarzzz:
# What 🕵️‍♀️

- changes the source of the proposer_fee_recipient field to bidTrace as opposed to the eth execution payload feeRecipient value.

# Why 🔑

- Currently when a builder with proposer payments sends blocks to our relay, it will now show the validators preferred fee recipient, but instead the fee recipient in the execution payload. This issue did not make itself known because the main builder being used with dreamboat in prod did not support proposer payments.

# How ⚙️
 - simple change of field source

# Testing 🧪


aratz-lasa (approved):
Nice!

PR #37, opened 2022-11-10
Reorg + fixes 

lukanus:


cmd/dreamboat/main.go
```diff
@@ -172,16 +198,105 @@ func setupKeys(c *cli.Context) (*blst.SecretKey, types.PublicKey, error) {
...
 
+		ds := datastore.NewDatastore(&datastore.TTLDatastoreBatcher{storage}, storage.DB)
+		if err = datastore.InitDatastoreMetrics(m); err != nil {
+			return err
+		}
+
+		regMgr := relay.NewProcessManager(c.Uint("relay-verify-queue-size"), c.Uint("relay-store-queue-size"))
+		regMgr.AttachMetrics(m)
+
+		reg, err := ds.GetAllRegistration()
```
aratz-lasa, line 241: Could we extract this into a method to make the `run` call more readable?

lukanus: To function of main package? like setupKeys above? Or where would you like to see it?

metrics/metrics.go
```diff
@@ -0,0 +1,53 @@
...
+	return promhttp.HandlerFor(
+		m.gatherers[0],
+		promhttp.HandlerOpts{
+			EnableOpenMetrics: true,
+		},
+	)
+}
+
+func AttachProfiler(sm *http.ServeMux) {
+	sm.HandleFunc("/debug/pprof/cmdline", pprof.Cmdline)
```
aratz-lasa, line 48: Iff we are using raw strings here, we should remove unnecessary pprof routes in `api.go`

pkg/relay/relay.go
```diff
@@ -0,0 +1,513 @@
...
+
+	PutHeader(context.Context, structs.Slot, structs.HeaderAndTrace, time.Duration) error
+	GetHeaders(context.Context, structs.Query) ([]structs.HeaderAndTrace, error)
+	GetMaxProfitHeadersDesc(context.Context, structs.Slot) ([]structs.HeaderAndTrace, error)
+
+	PutRegistrationRaw(context.Context, structs.PubKey, []byte, time.Duration) error
+	GetRegistration(context.Context, structs.PubKey) (types.SignedValidatorRegistration, error)
+}
+
+type RegistrationManager interface {
```
aratz-lasa, line 47: This interface is hard to understand. I think we should change the method names. Since in order to understand what they do, I must go to the implementation

pkg/relay/relay.go
```diff
@@ -0,0 +1,513 @@
...
+	rs.singleRetChannPool.Put(respCh)
+	timer2.ObserveDuration()
+	if resp.Err != nil {
+		logger.WithField(
+			"pubkey", proposerPubkey,
+		).Error("signature invalid")
+		return nil, fmt.Errorf("signature invalid")
+	}
+
+	timer3 := prometheus.NewTimer(rs.m.Timing.WithLabelValues("getPayload", "getPayload"))
```
aratz-lasa, line 239: Are we planning to leave the timers? Or is it just for performance analysis?

lukanus: I was suspicious, the times taking so long,  so I put them there. It wouldn't hurt to left it here, but totally fine to remove /rename them as well

pkg/relay/processmanager.go
```diff
@@ -0,0 +1,173 @@
...
+	"sync"
+	"time"
+
+	"github.com/blocknative/dreamboat/pkg/structs"
+	"github.com/flashbots/go-boost-utils/bls"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+type ProcessManager struct {
+	M   map[string]uint64
```
aratz-lasa, line 14: `M`, `acc` should have readable names, in order to understand their purpose. 

Also, what's the difference between Stack `A`, `B` and `C`? I think the name should make this clear. It's important variable naming in public repositories

aratz-lasa: If `A`, `B` and `C` are random channels, then we should put the explanation here (on top) instead of inside `VerifyParallel`

aratz-lasa: However, I think names should be something similar to `VerifySubmitBlockCh`, `VerifyRegisterValidatorCh` and `VerifyOtherCh`

lukanus: https://github.com/blocknative/dreamboat/pull/37/files/b46041f077c62cfb41dc90b9f26304def179d781#diff-0a50202f9351d5a88dd5428ef1ce9c8a264beff7ce7e7662f50b567dd630d073R142 

The names are A,B,C - as it's not processed in priority order, as you've said it's hard to say which one is really more important to process.  
The idea for that, was also so that package should be mostly unaware of what's inside and don't care about the payloads as every one is processed in the same way. I might have used something like containers.Ring, but this would have been more complicated. So kept this just as hardcoded select with those 3.

pkg/relay/register.go
```diff
@@ -0,0 +1,249 @@
...
+
+type Setter interface {
+	Set(k string, value uint64)
+}
+
+type Getter interface {
+	Get(k string) (value uint64, ok bool)
+}
+
+type SVRReq struct {
```
aratz-lasa, line 35: Could we rename this and the structs below to something more descriptive?

pkg/relay/register.go
```diff
@@ -0,0 +1,249 @@
...
+	}
+
+	respCh := rs.retChannPool.Get().(chan SVRReqResp)
+	defer rs.retChannPool.Put(respCh)
+
+	failure := make(chan struct{}, 1)
+	exit := make(chan error, 1)
+
+	sentVerified := uint32(0)
+	go registerSync(rs.regMngr, respCh, failure, exit, payload, &sentVerified)
```
aratz-lasa, line 140: Similarly here it is not clear why these 2 goroutines are spawned

aratz-lasa (changes requested):
These changes are amazing regarding performanec! Amazing work @lukanus 

There are 3 high-level things we must do before merging this PR:

- Refactor process manager and register logic, so that the execution flow is clear, as well as more decriptive variable names. 
- Run load tests in a testnet and make sure payloads are correctly delivered, validators are registered and blocks are submitted.
- Run load test in a testnet and extract official conclusions about the performance improvements

pkg/relay/processmanager.go
lukanus: I changed the name to your suggestions also M and acc to something more desciptive.

lukanus (commented):
A set of answeres to the requested changes

lukanus (commented):
A set of answeres to the requested changes

cmd/dreamboat/main.go
aratz-lasa: Yes, nice! But could we make it a method of `ProcessManager`?

pkg/relay/relay.go
aratz-lasa: Nice!

pkg/relay/relay.go
aratz-lasa: Okay, so it really depends on how much value these timers you consider they add. If you think they may be useful for future peformance improvements we leave them. Otherwise we just leave the outer timer. Your choice!

pkg/relay/register.go
aratz-lasa: Mmm I understand why you want to keeep it short. But VS could be anything. I think we should use VerifyReq, which is only one letter more than StoreReq that is below

cmd/dreamboat/main.go
lukanus: For processmanager - definitely not - as it doesn't really include the database. 
I can add this into the relay itself, however it may not necessarily be relay's area of interest, and separated from it. 
This one is only related to the process of initially loading this particular binary, that's why it's placed in the cmd folder and in neither of other modules.

pkg/relay/register.go
```diff
@@ -0,0 +1,250 @@
...
+
+type Getter interface {
+	Get(k string) (value uint64, ok bool)
+}
+
+// VSReq is a request structure used in communication
+// between api calls and fixed set of worker goroutines
+// it's using return channel pattern, meaning that after
+// sent the sender locks on that channel to get the response
+type VSReq struct {
```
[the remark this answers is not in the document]
lukanus: oks will do Verify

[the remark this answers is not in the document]
lukanus: done

dmarzzz: What does VS stand for?

metrics/metrics.go
```diff
@@ -0,0 +1,53 @@
...
+	reg.Register(collectors.NewGoCollector())
+
+	return &Metrics{registry: reg, gatherers: prometheus.Gatherers{reg}}
+}
+
+func (m *Metrics) RegisterDB(db *sql.DB, dbName string) error {
+	return m.registry.Register(collectors.NewDBStatsCollector(db, dbName))
+}
+
+func (m *Metrics) RegisterExpvar(exports map[string]*prometheus.Desc) error {
```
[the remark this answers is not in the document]
lukanus: https://pkg.go.dev/expvar 
It's needed for badger internal metrics

dmarzzz:
High level this is awesome great work !! It means a lot to me personally that were able to increase performance of the relay drastically while also keeping this application ~ simple ~ to run. My main concern is that it adds _**ALOT**_ of complexity to the codebase (often necessary) so I want to make sure we also prioritize readable and navigability in the code base. Since this is often hard to assess once you're in deep, a good test for this might be: take someone from Blocknative unfamiliar with the codebase and ask if them if they can verify what checks we run on validator registrations. 😃🚀

pkg/relay/old.go
```diff
@@ -0,0 +1,137 @@
+package relay
```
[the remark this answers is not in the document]
lukanus: I kept this for the sake of this PR (to be able to compare) also for benchmark, after everything would be fine, we can remove that. It's not used anywhere in application right now other then in the benchmarks.

pkg/relay/register.go
lukanus: Yeah I changed that already after Aratz's comment to VerifyReq :)

pkg/relay/register.go
```diff
@@ -0,0 +1,250 @@
...
+	defer rs.retChannPool.Put(respCh)
+
+	failure := make(chan struct{}, 1)
+	exit := make(chan error, 1)
+
+	sentVerified := uint32(0)
+	go registerSync(rs.regMngr, respCh, failure, exit, payload, &sentVerified)
+	go checkInMem(rs.beaconState, rs.regMngr, payload, respCh)
+	var failed bool
+	VInp := rs.regMngr.GetVerifyChan(ResponseQueueRegister)
```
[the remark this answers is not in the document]
lukanus: changed to verifyChan :)

aratz-lasa:
> High level this is awesome great work !! It means a lot to me personally that were able to increase performance of the relay drastically while also keeping this application ~ simple ~ to run. My main concern is that it adds _**A LOT**_ of complexity to the codebase (often necessary) so I want to make sure we also prioritize readable and navigability in the code base. Since this is often hard to assess once you're in deep, a good test for this might be: take someone from Blocknative unfamiliar with the codebase and ask if them if they can verify what checks we run on validator registrations. 😃🚀

I think a good idea is to ping @lthibault after @lukanus has finalised the refactoring. He would be the "Blocknative unfamiliar person"

cmd/dreamboat/main.go
```diff
@@ -193,18 +297,23 @@ func run() cli.ActionFunc {
...
 		// run the http server
 		g.Go(func() (err error) {
-			svr.BaseContext = func(l net.Listener) context.Context {
-				return ctx
+			svr = http.Server{
+				Addr:         c.String("addr"),
+				ReadTimeout:  c.Duration("timeout"),
+				WriteTimeout: c.Duration("timeout"),
+				IdleTimeout:  time.Second * 2,
+				BaseContext: func(l net.Listener) context.Context { // 99% not needed
```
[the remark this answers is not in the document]
lthibault: I think we should probably *not* replace the background context because it may interfere with graceful shutdown.

pkg/relay/processmanager.go
```diff
@@ -0,0 +1,178 @@
...
+	for i := uint(0); i < num; i++ {
+		go rm.ParallelStore(store, ttl)
+	}
+}
+
+func (rm *ProcessManager) RunCleanup(checkinterval uint64, cleanupInterval time.Duration) {
+	for {
+		now := uint64(time.Now().Unix())
+		var keys []string
+		rm.lrtl.RLock()
```
lthibault, line 62: Can we put both of these inner for-loops in subfunctions and use `defer` to unlock the mutex?  I am extremely paranoid about panics while holding locks, since HTTP servers will often recover from panicking requests in middleware and subsequently deadlock.

pkg/relay/processmanager.go
```diff
@@ -0,0 +1,178 @@
...
+		for k, v := range rm.LastRegTime {
+			if checkinterval < now-v {
+				keys = append(keys, k)
+			}
+		}
+		rm.lrtl.RUnlock()
+
+		rm.lrtl.Lock()
+		for _, k := range keys {
```
lthibault, line 71: (This is the second for loop in question.  Creating a comment so that we can explicitly resolve.)

pkg/relay/processmanager.go
```diff
@@ -0,0 +1,178 @@
...
+	rm.lrtl.RLock()
+	defer rm.lrtl.RUnlock()
+	value, ok = rm.LastRegTime[k]
+	return
+}
+
+func (rm *ProcessManager) ParallelStore(datas Datastore, ttl time.Duration) {
+	rm.m.RunningWorkers.WithLabelValues("ParallelStore").Inc()
+	defer rm.m.RunningWorkers.WithLabelValues("ParallelStore").Dec()
+	ctx := context.Background()
```
lthibault, line 128: Can we pass in a context instead of using `Background()` ?

pkg/relay/register.go
```diff
@@ -0,0 +1,296 @@
...
+
+	var respCh chan Resp
+
+	// This function is limitted to the size of the buffer to prevent deadlocks
+	if rs.config.RegisterValidatorMaxNum > uint64(len(payload)) {
+		respCh = rs.retChannPool.Get().(chan Resp)
+		defer rs.retChannPool.Put(respCh)
+	}
+
+	fc := NewFlowControl(respCh, len(payload))
```
lthibault, line 71: Can we add a comment clarifying that it's OK for `respCh` to be `nil`?  I had to double-check the implementation of `NewFlowControl` to be certain.  Additionally, can we add a docstring to `NewFlowControl` that declares a contract for how nil channels are handled?

pkg/relay/register.go
```diff
@@ -0,0 +1,296 @@
...
+	go synchAndStore(rs.regMngr, fc, payload)
+
+	verifyChan := rs.regMngr.GetVerifyChan(ResponseQueueRegister)
+SendPayloads:
+	for i, p := range payload {
+		msg, err := types.ComputeSigningRoot(payload[i].Message, rs.config.BuilderSigningDomain)
+		if err != nil {
+			fc.SentVerificationInc()
+			fc.RespCh <- Resp{ID: i, Err: errors.New("invalid signature"), Type: ResponseTypeVerify}
+			break SendPayloads
```
lthibault, line 90: I believe the label is unnecessary here.

pkg/relay/register.go
```diff
@@ -0,0 +1,296 @@
...
+		}
+
+		select {
```
lthibault, line 93: Can we put this into a function and return a bool that determines whether to break out of the loop?  Labeled breaks are hard to follow, and the statefulness tends to cause bugs.

pkg/relay/register.go
```diff
@@ -0,0 +1,296 @@
...
+	respCh := rs.singleRetChannPool.Get().(chan Resp)
+	defer rs.singleRetChannPool.Put(respCh)
+
+	rs.regMngr.GetVerifyChan(ResponseQueueRegister) <- VerifyReq{
+		Signature: payload.Signature,
+		Pubkey:    payload.Message.Pubkey,
+		Msg:       msg,
+		Response:  respCh,
+	}
+	r := <-respCh
```
lthibault, line 135: Is this guaranteed not to block?  Can we add a comment explaning why?

Either way, can we select along with `<-ctx.Done()` just to be safe?

pkg/relay/register.go
```diff
@@ -0,0 +1,296 @@
...
+	timer2 := prometheus.NewTimer(rs.m.Timing.WithLabelValues("registerValidatorSingular", "verify"))
+	msg, err := types.ComputeSigningRoot(payload.Message, rs.config.BuilderSigningDomain)
+	if err != nil {
+		return errors.New("invalid signature")
+	}
+
+	respCh := rs.singleRetChannPool.Get().(chan Resp)
+	defer rs.singleRetChannPool.Put(respCh)
+
+	rs.regMngr.GetVerifyChan(ResponseQueueRegister) <- VerifyReq{
```
lthibault, line 129: `<paranoia>`
When is the verification channel closed?  Are we certain we never return a closed or nil channel?
`</paranoia>`

pkg/relay/register.go
```diff
@@ -0,0 +1,296 @@
...
+	}
+	r := <-respCh
+	timer2.ObserveDuration()
+
+	if r.Err != nil {
+		return errors.New("invalid signature")
+	}
+
+	timer3 := prometheus.NewTimer(rs.m.Timing.WithLabelValues("registerValidatorSingular", "store"))
+	storeCh := rs.regMngr.GetStoreChan()
```
lthibault, line 143: Same questions as above re: `storeCh`.

pkg/relay/verify.go
```diff
@@ -0,0 +1,54 @@
...
+	"github.com/flashbots/go-boost-utils/types"
+	blst "github.com/supranational/blst/bindings/go"
+)
+
+var dst = []byte("BLS_SIG_BLS12381G2_XMD:SHA-256_SSWU_RO_POP_")
+
+const BLST_SUCCESS = 0x0
+
+func VerifySignatureBytes(msg [32]byte, sigBytes, pkBytes []byte) (ok bool, err error) {
+	defer func() { // better safe than sorry
```
lthibault, line 16: `// better safe than sorry`

Hell yeah, brother!

lthibault (changes requested):
Looks good overall.  Inline comments fall into roughly two categories:

1. Proposed changes that focus on defensive programming (i.e. not a bug, but "better safe than sorry")
2. Questions in need of clarification or double-checking

cmd/dreamboat/main.go
```diff
@@ -172,16 +198,94 @@ func setupKeys(c *cli.Context) (*blst.SecretKey, types.PublicKey, error) {
...
+		timeDataStoreStart := time.Now()
+		m := metrics.NewMetrics()
+
+		storage, err := badger.NewDatastore(config.Datadir, &badger.DefaultOptions)
+		if err != nil {
+			config.Log.WithError(err).Fatal("failed to initialize datastore")
+			return err
+		}
+		config.Log.
+			WithFields(logrus.Fields{
```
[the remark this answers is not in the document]
lukanus: 👍

pkg/relay/register.go
lukanus: Never, it's shared over the fix set of workers, that are started with the binary.

pkg/relay/register.go
lukanus: same as above - "Never, it's shared over the fix set of workers, that are started with the binary. "

pkg/relay/register.go
lukanus: It's not necessary here, but since it is down below. 
I added it here as well, for visibility.

pkg/relay/register.go
lukanus: I'm not exactly following what kind of change are you talking bout here :) 
This operation has to be a channel select.

lthibault: I'm proposing that we wrap the entire select statement in its own function, which would return a bool that indicates whether to break out of the loop.  The idea is to get rid of the `SendPayloads` label.

pkg/relay/register.go
lthibault: Makes sense.  SGTM.  Maybe add a small comment saying something like `// guaranteed to receive` ?

pkg/relay/register.go
lthibault: I don't like labels 😉

pkg/relay/processmanager.go
```diff
@@ -0,0 +1,178 @@
...
+		return rm.VerifyRegisterValidatorCh
+	default: // ResponseQueueOther
+		return rm.VerifyOtherCh
+	}
+}
+
+func (rm *ProcessManager) Set(k string, value uint64) {
+	rm.lrtl.Lock()
+	defer rm.lrtl.Unlock()
+	defer rm.m.MapSize.Set(float64(len(rm.LastRegTime)))
```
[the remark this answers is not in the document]
lthibault: Haha, yeah why not 😄.  "Principle of least surprise" and all that...

metrics/metrics.go
```diff
@@ -0,0 +1,53 @@
...
+	return m.registry.Register(collectors.NewExpvarCollector(exports))
+}
+
+func (m *Metrics) Register(cs prometheus.Collector) error {
+	return m.registry.Register(cs)
+}
+
+func (m *Metrics) Handler() http.Handler {
+	return promhttp.HandlerFor(
+		m.gatherers[0],
```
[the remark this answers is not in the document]
lukanus: There is one that we're registering everything to in the New
```golang
	reg := prometheus.NewRegistry()
	reg.MustRegister(collectors.NewProcessCollector(collectors.ProcessCollectorOpts{}))
	reg.Register(collectors.NewGoCollector())
```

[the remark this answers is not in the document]
lthibault: 👍 Can we add a comment here, and at the `Register` call that explains the relation?  I want to hedge against the possibility that we later decide to remove the `Register` call and cause a downstream panic.  Something like `// there MUST be at least one collector registered` and `// New always registers a collector` would be perfect.

pkg/relay/processmanager.go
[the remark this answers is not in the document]
lukanus: Changed

pkg/relay/processmanager.go
lukanus: It's simple append here but well, as you wish... :)

lukanus: done

pkg/relay/processmanager.go
lukanus: done

pkg/relay/processmanager.go
lthibault: Yeah, I hesitated to even mention it, but it helps to not stop and check while re-reading the code.  So if nothing else, I think it's helpful to signal quite loudly that the mutex behavior is correct.  🤷‍♂️ 

Anyway, thanks :)

cmd/dreamboat/main.go
```diff
@@ -193,18 +297,23 @@ func run() cli.ActionFunc {
...
+			svr = http.Server{
+				Addr:         c.String("addr"),
+				ReadTimeout:  c.Duration("timeout"),
+				WriteTimeout: c.Duration("timeout"),
+				IdleTimeout:  time.Second * 2,
+				BaseContext: func(l net.Listener) context.Context { // 99% not needed
+					return ctx
+				},
+				Handler:        mux,
+				MaxHeaderBytes: 4096,
```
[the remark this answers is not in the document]
lukanus: It's just moved - haven't changed that and the context above.  Let's not touch this and other timeouts for now, and iterate on this later. As I'm hoping this main function become more... vanilla.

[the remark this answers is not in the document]
lthibault: Sounds reasonable 👍

metrics/metrics.go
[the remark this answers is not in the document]
lukanus: I think as a rule of thumb if the structure has constructor you have to use this. Unless you really understand what are you doing. So similarly in a lot of different places you may have expect nils if you don't use a constructor.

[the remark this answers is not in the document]
lthibault: Yeah, I guess that's fair.  Carry on :)

cmd/dreamboat/main.go
```diff
@@ -172,16 +198,94 @@ func setupKeys(c *cli.Context) (*blst.SecretKey, types.PublicKey, error) {
...
+
+		api := api.NewApi(config.Log, service)
+		api.AttachMetrics(m)
+
+		regMgr.RunStore(ds, config.TTL, c.Uint("relay-workers-store-validator"))
+		regMgr.RunVerify(c.Uint("relay-workers-verify"))
+
+		config.Log.WithFields(logrus.Fields{
+			"service":     "relay",
+			"startTimeMs": time.Since(timeRelayStart).Milliseconds(),
```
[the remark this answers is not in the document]
lukanus: Let's not touch it this for now, as it was copypasted

cmd/dreamboat/main.go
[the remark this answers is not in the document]
lukanus: Yes, but as this needs more investigation and this is what was there before - Let's not touch it :)

cmd/dreamboat/main.go
```diff
@@ -172,16 +198,94 @@ func setupKeys(c *cli.Context) (*blst.SecretKey, types.PublicKey, error) {
...
+		if err != nil {
+			return err
+		}
+
+		timeDataStoreStart := time.Now()
+		m := metrics.NewMetrics()
+
+		storage, err := badger.NewDatastore(config.Datadir, &badger.DefaultOptions)
+		if err != nil {
+			config.Log.WithError(err).Fatal("failed to initialize datastore")
```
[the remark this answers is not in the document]
lukanus: I'll decrease this level for now untill we wouldn't go out from this cli package

cmd/dreamboat/main.go
```diff
@@ -172,16 +198,94 @@ func setupKeys(c *cli.Context) (*blst.SecretKey, types.PublicKey, error) {
...
+		service := pkg.NewService(config.Log, config, ds, r, as)
+		service.AttachMetrics(m)
+
+		api := api.NewApi(config.Log, service)
+		api.AttachMetrics(m)
+
+		regMgr.RunStore(ds, config.TTL, c.Uint("relay-workers-store-validator"))
+		regMgr.RunVerify(c.Uint("relay-workers-verify"))
+
+		config.Log.WithFields(logrus.Fields{
```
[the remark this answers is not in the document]
lukanus: changed

cmd/dreamboat/main.go
```diff
@@ -172,16 +198,94 @@ func setupKeys(c *cli.Context) (*blst.SecretKey, types.PublicKey, error) {
...
+
+		storage, err := badger.NewDatastore(config.Datadir, &badger.DefaultOptions)
+		if err != nil {
+			config.Log.WithError(err).Fatal("failed to initialize datastore")
+			return err
+		}
+		config.Log.
+			WithFields(logrus.Fields{
+				"service":     "datastore",
+				"startTimeMs": time.Since(timeDataStoreStart).Milliseconds(),
```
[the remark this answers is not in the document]
lukanus: I'll keep this duration as it is as it was

cmd/dreamboat/main.go
```diff
@@ -93,16 +104,40 @@ var flags = []cli.Flag{
 		Value:   24 * time.Hour,
 		EnvVars: []string{"BN_RELAY_TTL"},
 	},
-	&cli.BoolFlag{
-		Name:  "checkKnownValidator",
-		Usage: "rejects validator registration if it's not a known validator from the beacon",
-		Value: false,
+	&cli.Uint64Flag{
+		Name:    "relay-max-request",
```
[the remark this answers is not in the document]
lukanus: changed that to be different parameter after slack convo

cmd/dreamboat/main.go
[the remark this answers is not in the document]
lukanus: changed

cmd/dreamboat/main.go
```diff
@@ -227,6 +336,22 @@ func run() cli.ActionFunc {
...
 
+func loadRegistrations(ds *datastore.Datastore, regMgr *relay.ProcessManager) {
+	reg, err := ds.GetAllRegistration()
+	if err == nil {
+		for k, v := range reg {
+			regMgr.Set(k, v.Message.Timestamp)
+		}
+
+		config.Log.
+			WithFields(logrus.Fields{
```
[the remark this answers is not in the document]
lukanus: done

pkg/relay/processmanager.go
lukanus: This context is local to the workers, and as there is no control on that so it's kind of done on purpose

pkg/relay/register.go
lukanus: Added comment

lthibault: Nevermind, I see it now.  GitHub is so weird sometimes...

dmarzzz (approved):

pkg/relay/register.go
```diff
@@ -102,7 +106,14 @@ SendPayloads:
 			fc.SentVerificationInc()
 		}
 	}
-	return <-fc.ExitCh
+
+	err := <-fc.ExitCh
```
aratz-lasa, line 110: Is this expected to always receive? If so I'll add a comment.

lukanus: yes, otherwise nothing would ever finish :)

aratz-lasa (approved):
Ready for merge!

PR #38, opened 2022-11-11
Add timestamp validation on block submission

aratz-lasa:
# What 🕵️‍♀️
It is necessary to check whether the timestamp of the block submissions is the one expected for the corresponding slot. 

# Why 🔑
If validation is not implemented, then (correctly implemented) validators to reject the blocks the relay serves them.

# How ⚙️
Obtain Genesis time on relay initialization and compute expected timestamp on block submissions by doing `genesisTime + slot x 12`

# Testing 🧪
- [x] add tests to check timestamp validation
- [x] end-to-end testing in testnet 

# Special mentions
Thank you to @metachris for notifying about this, and help make the relay more robust.

### DIFF
--- a/internal/mock/pkg/beacon.go
+++ b/internal/mock/pkg/beacon.go
@@ -50,10 +50,10 @@ func (mr *MockBeaconClientMockRecorder) Endpoint() *gomock.Call {
 }
 
 // Genesis mocks base method.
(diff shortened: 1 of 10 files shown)


dmarzzz (approved):
LGTM 👍

PR #39, opened 2022-11-11
Fix/prevrandao

aratz-lasa:
# What 🕵️‍♀️
Implements randao validation on block submissions

# Why 🔑
Otherwise it may cause security vulnerabilities

# How ⚙️
Query the randao from the beacon whenever the headslot is updated and check for validation on block submission

# Testing 🧪
- [x] add unit test to check for invalid randao
- [ ] end-to-end testing in testnet


lukanus:
Implemented in shapella branch :)

PR #40, opened 2022-11-14
Set JSON timestamp to RFC3339Nano.

lthibault:
Adds nanosecond precision to logs (JSON format only).

cc @bkellerman: for context, and in case you ever need to fix this somewhere else

### DIFF
--- a/cmd/dreamboat/main.go
+++ b/cmd/dreamboat/main.go
@@ -101,7 +101,7 @@ var flags = []cli.Flag{
 }
 
 var (
-	config = relay.Config{Log: log.New()}
+	config relay.Config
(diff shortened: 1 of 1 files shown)


dmarzzz (approved):
LGTM - my gut reaction is to say the time unit should be configurable but I believe its safe to assume users of the app want highest precision

lthibault:
> LGTM - my gut reaction is to say the time unit should be configurable but I believe its safe to assume users of the app want highest precision

It's ghetto-configurable in the sense that you can always create your own `main.go` file that imports dreamboat, and change the JSON formatter. 🤷‍♂️

PR #41, opened 2022-11-16
fix proposerFeeRecipient in payload deliveries

aratz-lasa:
# What 🕵️‍♀️
Fixes the FeeRecipient field on delivered payloads Data API.

# Why 🔑
Otherwise Data API consumers may be confused on who receives the MEV

# How ⚙️
Change the delivered payload trace to use the submitted block fee recipient

# Testing 🧪

### DIFF
--- a/pkg/service.go
+++ b/pkg/service.go
@@ -17,7 +17,7 @@ import (
 )
 
 const (
-	Version = "0.2.5"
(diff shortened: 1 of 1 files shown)


bkellerman (approved):
LGTM

lukanus (approved):

PR #42, opened 2022-11-16
Change put header logic in block submissions

lukanus:
# What 🕵️‍♀️
Removes Badger reads and writes in favor of local map cache.
Changes badger write strategy - Submissions are written into database as separate records, then after every X minutes/slots data is persisted for entire slot. Upon failure and restart process is reading all orphan nodes from badger, and resubmits it under one key. 
Where possible, the singular key is now used to read data. 
Some Database interfaces got narrowed to represent parameters they take, as it was not possible to pass multiple parameters in the Query structure before. 
Database controller structure got extended as datastore doesn't allow to make separate transactions and prefix scans.


pkg/datastore/headers.go
```diff
@@ -0,0 +1,433 @@
...
+	// check and load keys if exists
+	return s.loadKeysAndCleanup(ctx, uint64(hr.Slot))
+}
+
+func (s *Datastore) loadKeysAndCleanup(ctx context.Context, slot uint64) error {
+	s.l.Lock()
+	defer s.l.Unlock()
+
+	// need to load keys to memory
+	data, err := s.TTLStorage.Get(ctx, HeaderKey(slot))
```
aratz-lasa, line 150: Is there an easy way to avoid doing a `Get` call to the DB in every `PutHeader`?

pkg/datastore/headers.go
```diff
@@ -0,0 +1,433 @@
...
+	slot, err := s.TTLStorage.Get(ctx, HeaderHashKey(hash))
+	if err != nil {
+		return nil, err
+	}
+	return s.GetHeadersBySlot(ctx, binary.LittleEndian.Uint64(slot))
+}
+
+func (s *Datastore) GetLatestHeaders(ctx context.Context, limit uint64) ([]structs.HeaderAndTrace, error) {
+	ls := s.hc.GetLatestSlot()
+	stop := ls - uint64(24*time.Hour/time.Second*12)
```
aratz-lasa, line 226: Here we should add a check for underflow

pkg/datastore/headers.go
```diff
@@ -0,0 +1,433 @@
...
+			return nil, err
+		}
+		return []structs.HeaderAndTrace{header}, nil
+	}
+	return headers, nil
+}
+
+// SaveHeaders is meant to persist the all the keys under one key
+// As optimization in future this function can operate only on database, so instead from memory it may just reorganize keys
+func (s *Datastore) SaveHeaders(ctx context.Context, slots []uint64, ttl time.Duration) error {
```
aratz-lasa, line 283: Why not `StoreHeaders` for consistency

pkg/datastore/headers.go
```diff
@@ -0,0 +1,433 @@
...
+	for _, slot := range slots {
+		if err := s.saveHeader(ctx, slot, ttl); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (s *Datastore) saveHeader(ctx context.Context, slot uint64, ttl time.Duration) error {
```
aratz-lasa, line 292: Same for this and following functions, why not `store`, instead of `save`?

pkg/datastore/headers.go
```diff
@@ -0,0 +1,433 @@
...
+	return nil
+}
+
+type LoadItem struct {
+	Time    uint64
+	Content []byte
+}
+
+// FixOrphanHeaders is reading all the orphan headers from
+func (s *Datastore) FixOrphanHeaders(ctx context.Context, ttl time.Duration) error {
```
aratz-lasa, line 343: We should add unit tests for this function. Since some time ago Badgerdb had some issues/bugs for iterating based on a prefix. But probably the bug is fixed

pkg/datastore/headers.go
```diff
@@ -0,0 +1,433 @@
...
+
+func (s *Datastore) MemoryCleanup(ctx context.Context, ttl time.Duration) error {
+	for {
+		time.Sleep(InMemorySlotPurgeCheck)
+		slots, ok := s.hc.CheckForRemoval()
+		if !ok {
+			continue
+		}
+
+		s.SaveHeaders(ctx, slots, ttl)
```
aratz-lasa, line 430: After storing headers, where are the memory items being removed?

cmd/dreamboat/main.go
```diff
@@ -227,11 +227,19 @@ func run() cli.ActionFunc {
 
 		timeRelayStart := time.Now()
 		as := &pkg.AtomicState{}
-		ds := datastore.NewDatastore(&datastore.TTLDatastoreBatcher{storage}, storage.DB)
+
+		hc := datastore.NewHeaderController()
```
aratz-lasa, line 231: This seems something that should be created within the function `NewDatastore`, instead of out here. It is not used anywhere else except from inside the `Datastore`.

pkg/datastore/headerscontroller.go
```diff
@@ -0,0 +1,282 @@
...
+
+	s, ok := hc.content[slot]
+	if !ok {
+		return hnt, false
+	}
+	return s.GetMaxProfit()
+
+}
+
+type HNTs struct {
```
aratz-lasa, line 203: What does HNTs mean? We should use more descriptive name for this

pkg/datastore/headerscontroller.go
```diff
@@ -0,0 +1,282 @@
...
+	S StoredIndex
+
+	rev                        uint64
```
aratz-lasa, line 206: What does `rev` mean?

pkg/datastore/headerscontroller.go
```diff
@@ -0,0 +1,282 @@
...
+	hc.cl.RUnlock()
+
+	if !ok {
+		return nil, 0, nil
+	}
+	elements, revision = h.GetContent()
+	return elements, revision, err
+}
+
+func (hc *HeaderController) UnlinkElement(slot, expectedRevision uint64) (success bool) {
```
aratz-lasa, line 161: `UnlinkElement` is accurate, but for giving a more descriptive name in this context, I would call it something similar to `deleteFromMemory`

pkg/datastore/headerscontroller.go
```diff
@@ -0,0 +1,282 @@
...
+
+		ls := atomic.LoadUint64(hc.latestSlot)
+		if ls == 0 || slot > ls {
+			atomic.StoreUint64(hc.latestSlot, slot)
+		}
+	}
+	return newCreated, h.AddContent(hnt)
+}
+
+func (hc *HeaderController) GetContent(startingSlot, stopSlot uint64, limit int) (elements []structs.HeaderAndTrace, lastSlot uint64) {
```
aratz-lasa, line 127: Could we change `content` for `headers`? Here and in the struct name. 

Since the HeaderController, should be working with `headers`, and it's more readable I think.

pkg/datastore/headerscontroller.go
```diff
@@ -0,0 +1,282 @@
...
+		}
+		toBeRemoved = append(toBeRemoved, v.Slot)
+		ok = true
+	}
+
+	return toBeRemoved, ok
+
+}
+
+func (hc *HeaderController) GetOrderedDesc() (info []uint64) {
```
aratz-lasa, line 48: This should be lower case since it's not used outside `HeadController`

pkg/structs/structs.go
```diff
@@ -182,3 +201,20 @@ type GenesisInfo struct {
 	GenesisValidatorsRoot string `json:"genesis_validators_root"`
 	GenesisForkVersion    string `json:"genesis_fork_version"`
 }
+
+// / extra
+type HR struct {
```
aratz-lasa, line 206: What does HR mean? We should use a more descriptive name

aratz-lasa (changes requested):
Nice wok @lukanus !

I set a couple of comments regarding readability, and general question! But I think it's really good

cmd/dreamboat/main.go
lukanus: Well, I disagree. As much as we can we should avoid any composition and initialization within constructors, as a rule of thumb. I'm happy to make it an interface within datastore.

pkg/datastore/headers.go
lukanus: It's called only upon the very first payload to the new slot.
It's needed in the scenario when someone would like to add data into the slot that's not in memory. So it loads it on demand (if needed)

pkg/datastore/headers.go
lukanus: I wanted to distinct it a little bit from the store event. 
As this is meant for a scenario when go you from memory into the database.
It's not a big deal to rename it from Save but I think it would need to be something more descriptive than Store

pkg/datastore/headers.go
lukanus: https://github.com/blocknative/dreamboat/blob/0b5d6988e24d15955eb28ab95ed24b1e74d70c5e/pkg/datastore/headers.go#L302

pkg/datastore/headerscontroller.go
lukanus: Delete from memory happens later by GC. This is just unlinking this from the map. Maybe we can find some better name...

pkg/datastore/headerscontroller.go
lukanus: oks added

pkg/datastore/headerscontroller.go
lukanus: Revision of this record. So in case any addition would happen mid unlinking process - we won't persist the data till next cycle.

pkg/structs/structs.go
lukanus: It deviated from HeaderRequest. but I think HeaderData would be fine...

lukanus: done

pkg/datastore/headers.go
lukanus: hm? It's exactly the same way how it was no ? :) You're thinking of math.Floor()?

aratz-lasa: Okay, don't worry, we can fix this in another PR. There is a PR opened that fixes this. But not for this PR.

pkg/datastore/headerscontroller.go
aratz-lasa: Yes, yes, delete is not accurate enough. What about something like `UnmapHeader`, or something on that direction?

pkg/datastore/headerscontroller.go
lukanus: "Header And Trace" -> "Header'N'Trace" -> "HNT" . There are a lot wrappers of the basic structure. 
I didn't have a better idea. `SlotContents` ?

pkg/datastore/headerscontroller.go
lukanus: `RemoveSlot` ?

cmd/dreamboat/main.go
aratz-lasa: Okay,  good point.The thing is that the more components the relay has, the bigger the main will become. So, start extracting these initializations into functions, for readability. So in main we should have a function call for every major component initialization:
`initDatastore`: initializes badger, headerController and datastore. Returns datastore
`initRelay`: initializes processManager and relay. Returns relay

What do you think about this alternative?

pkg/datastore/headerscontroller.go
aratz-lasa: Good for me

cmd/dreamboat/main.go
lukanus: I would very much like to change this function, just didn't wanted to do this in this very PR :) 
As we can clean main up from using errgroup and other things we don't need. 
But for that there are some changes needed to Service and Becon that I think should be done prior to cleanup :)

pkg/datastore/headerscontroller.go
lukanus: done

pkg/datastore/headerscontroller.go
aratz-lasa: Okay, I see, but in the following code: 
```
if expectedRevision != s.GetRevision() {
		return false
	}

	// we're only unlinking from map here
	// it should be later on eligible for GC
	delete(hc.content, slot)

```
Isn't possible to update the slot after checking revision, but before deleting the slot?

pkg/datastore/headers.go
aratz-lasa: Aaah I missed the newlyCreated conditional on `PutHeader`. Nice 👍

cmd/dreamboat/main.go
aratz-lasa: Perfect, then let's improve that in another PR

pkg/datastore/headers.go
```diff
@@ -0,0 +1,433 @@
...
+	}
+}
+
+type IndexEl struct {
+	Hash          [32]byte
+	Value         *big.Int
+	BuilderPubkey [48]byte
+}
+
+type Info struct {
```
aratz-lasa, line 71: What about `SlotInfo`?

pkg/datastore/headers.go
```diff
@@ -0,0 +1,433 @@
...
+	SubmissionsByPubKeys map[[48]byte]IndexEl
+}
+
+func NewStoreIndex() StoredIndex {
+	return StoredIndex{
+		SubmissionsByPubKeys: make(map[[48]byte]IndexEl),
+	}
+}
+
+type IndexEl struct {
```
aratz-lasa, line 65: What does `El` stand for?

lukanus: Element? It's meant to be shorter

pkg/datastore/headerscontroller.go
aratz-lasa: What about `OptimizedHeaders` or `IndexedHeaders`?

Also why do you use `uint64` instead of `Slot`? Any reason?

pkg/datastore/headers.go
lukanus: Changed

pkg/datastore/headerscontroller.go
```diff
@@ -0,0 +1,282 @@
...
+	"sort"
+	"sync"
+	"sync/atomic"
+	"time"
+
+	"github.com/blocknative/dreamboat/pkg/structs"
+)
+
+type HeaderController struct {
+	content map[uint64]*HNTs
```
aratz-lasa, line 13: Could we replace `content` by `headers`  in this and the other structs? In order to make what the content is clear

pkg/datastore/headerscontroller.go
lukanus: As much as I see the only place where it has anything called from it is `Epoch()` method  inside Service.
It's never really used in any operation, and it's simply just a number. Especially as we grow more into modules, it would be significantly easier to operate on base types, as this doesn't carry any additional value with it. 

`IndexedHeaders` will do.

pkg/datastore/headerscontroller.go
lukanus: done

pkg/datastore/headers.go
lukanus: It's the element of the index, so if we renamed is as Stored it should be rather StoredIndexElement.  but it's quite long

pkg/datastore/headerscontroller.go
lukanus: renamed

pkg/datastore/headers.go
aratz-lasa: Okay I see, what about `IndexMeta`? Since it's the metadata about the index, and `El` is not intuitively related to element

lukanus: Added as IndexMeta

aratz-lasa (approved):
Amazing work @lukanus ! Once we run end-to-end tests and load tests, we should merge and create a release of this and #37

PR #43, opened 2022-11-18
verify fee recipient of submitted block

dmarzzz:
# What 🕵️‍♀️

Adds a check to verify that a block submissions proposer fee recipient matches the fee recipient validators have registered. Thanks to @metachris for pointing this out!

# Why 🔑

Currently dreamboat does not support external builders, but instead a loosely defined trusted builder. We are in the process of hardening to be able to support any external builder with hopefully 0 trust.

# How ⚙️

On block submission we grab the validators registered fee recipient and compare it to the value in the block.

# Testing 🧪

### DIFF
--- a/pkg/relay.go
+++ b/pkg/relay.go
@@ -539,7 +539,7 @@ func (rs *DefaultRelay) verifyBlock(ctx context.Context, SubmitBlockRequest *typ
 	reg, err := state.Datastore().GetRegistration(ctx, pubKey)
 	if err != nil {
 		if !errors.Is(err, ds.ErrNotFound) {
-			rs.Log().Warn("unexpected datastore error")
+			rs.Log().WithError(err).Warn("unexpected datastore error")
 		}
 		return false, ErrUnregisteredSlot
 	}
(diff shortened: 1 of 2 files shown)


pkg/relay.go
```diff
@@ -526,11 +528,26 @@ func (rs *DefaultRelay) GetValidators(state State) BuilderGetValidatorsResponseE
...
-func (rs *DefaultRelay) verifyBlock(SubmitBlockRequest *types.BuilderSubmitBlockRequest) (bool, error) {
+func (rs *DefaultRelay) verifyBlock(ctx context.Context, SubmitBlockRequest *types.BuilderSubmitBlockRequest, state State) (bool, error) {
+
 	if SubmitBlockRequest == nil {
 		return false, fmt.Errorf("block empty")
 	}
 
+	pubKey := PubKey{SubmitBlockRequest.Message.ProposerPubkey}
+
+	reg, err := state.Datastore().GetRegistration(ctx, pubKey)
```
aratz-lasa, line 539: This is quite inefficient in the current setting, since it requires a DB io, in every block submission. Instead, we should keep this information in memory every epoch / slot. But let's wait for https://github.com/blocknative/dreamboat/pull/37 before implementing this.

dmarzzz: Great idea!

lukanus:
That one is implemented some time ago so, closing this one :)

PR #44, opened 2022-11-19
fix submitBlock default error

dmarzzz:
# What 🕵️‍♀️

"Manually fuzz testing" and noticed submitBlock returns errors like:
```
{"code":400,"message":"EOF"}
```
when it should be 
```
{"code":400,"message":"EOF"}
```
like the other endpoints

# Why 🔑

API consistency

# Testing 🧪


lukanus (approved):

lukanus:
@dmarzzz  I rebased that on top of current version (the file from your PR no longer existed :) ) And I'll merge this into the current awaiting set of changes/fixes - so it'll be released in the next occasion :)

PR #51, opened 2022-11-23
Return error properly, add Header metrics

lukanus:


pkg/datastore/headerscontroller.go
```diff
@@ -18,17 +18,20 @@ type HeaderController struct {
...
+	hc := &HeaderController{
 		slotLag:     slotLag,
 		slotTimeLag: slotTimeLag,
-
-		latestSlot: &latestSlot,
-		headers:    make(map[uint64]*IndexedHeaders),
+		latestSlot:  &latestSlot,
+		headers:     make(map[uint64]*IndexedHeaders),
 	}
+	hc.initMetrics()
```
aratz-lasa, line 33: the initMetrics was not called before. Why do we call it no?

aratz-lasa (changes requested):
Could we change `log` for `logger` in: https://github.com/blocknative/dreamboat/blob/73c78004fc292f84735699c56fedc49f076571e4/pkg/relay/relay.go#L160

pkg/datastore/headerscontroller.go
lukanus: The other ones are static, it's initiated as in the other ones

aratz-lasa (approved):
LGTM

PR #53, opened 2022-11-25
Feature lazy-store

lukanus:


aratz-lasa (changes requested):
Looks good! I added a couple of comments related to 2 main things:

1. Make `SendStore` fully async, avoiding getting blocked due to locks
2. Refactor struct names

aratz-lasa (approved):
LGTM!

Before merging we should test it in a testnet, but after that is ready.

PR #55, opened 2022-12-09
Api change

lukanus:
# What 🕵️‍♀️
Removes service - relay dependency

# Why 🔑
Needed for incremental set of changes for badgerdb migration.

# How ⚙️
API - related endpoints in service were moved into relay.  for further change in next PRs


cmd/dreamboat/main.go
```diff
@@ -308,10 +308,10 @@ func run() cli.ActionFunc {
 		}, as, ds, regMgr, auctioneer)
 		r.AttachMetrics(m)
 
-		service := pkg.NewService(config.Log, config, ds, r, as)
+		service := pkg.NewService(config.Log, config, ds, as)
 		service.AttachMetrics(m)
 
-		api := api.NewApi(config.Log, service)
+		api := api.NewApi(config.Log, r)
```
jcorry, line 314: Using `api` as an identifier shadows the `api` package name, which can be confusing at best and produce unwanted results at worst (ie: you can no longer call package level functions in the `api` package)

`api` is a pretty perfect identifier name for your API. The only problem with that name is the conflict with the package name (which is also well named). For clarity, aliasing the package to `dreamboatapi` so that you can call `dreamboatapi.NewApi()` might be an improvement. That's a weird thing to say though given that a reasonable rule of thumb is that if you are aliasing your own packages, that's a pretty clear indication you've named something poorly.

lukanus: :) it was just regular bug - my bad, don't worry I'll change the variable name to be different. :) Also I'm definitely against renaming the package to a very long identifier

jcorry: Yeah I definitely wouldn't rename the package. I am conflicted here because `api` is clearly the best name for the package...and `api` is also a very natural name for the value returned by `api.NewApi()` (I mean...obviously, right?)

lukanus: :) I think from the perspective of the future growth the better name for the package that have NewAPI function would probably be httpTransport or something like that. but right now I'd go minimalistic

aratz-lasa (approved):
LGTM

PR #56, opened 2022-12-12
Feature/immediate last payloads 3

aratz-lasa:


pkg/datastore/datastore.go
```diff
@@ -32,19 +35,30 @@ type Badger interface {
 }
 
 type Datastore struct {
+	Logger log.Logger
 	TTLStorage
 	Badger
+	*Auctioneer
```
lukanus, line 41: give it a name, like "a" or something

PR #57, opened 2022-12-12
Immediate last payloads

lukanus:
# What 🕵️‍♀️
Add memory caching for getting max / latest header and payload 

# Why 🔑
To increase relay performance for critical paths.

### DIFF
--- a/pkg/auction/auctioneer.go
+++ b/pkg/auction/auctioneer.go
@@ -3,62 +3,79 @@ package auction
 import (
 	"sync"
 
+	"github.com/lthibault/log"
+
 	"github.com/blocknative/dreamboat/pkg/structs"
 	"github.com/flashbots/go-boost-utils/types"
 )
 
 type Auctioneer struct {
-	mu                   sync.RWMutex
-	maxProfit            *structs.CompleteBlockstruct
-	latestBlockByBuilder map[types.PublicKey]*structs.CompleteBlockstruct
+	l log.Logger
+
+	mu                   [2]sync.RWMutex
+	maxProfit            [2]*structs.CompleteBlockstruct
+	latestBlockByBuilder [2]map[types.PublicKey]*structs.CompleteBlockstruct
 }
 
-func NewAuctioneer() *Auctioneer {
+func NewAuctioneer(l log.Logger) *Auctioneer {
 	return &Auctioneer{
-		latestBlockByBuilder: make(map[types.PublicKey]*structs.CompleteBlockstruct),
+		l:                    l.WithField("relay-service", "auctioneer"),
+		mu:                   [2]sync.RWMutex{},
+		maxProfit:            [2]*structs.CompleteBlockstruct{},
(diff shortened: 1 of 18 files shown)


pkg/datastore/datastore.go
```diff
@@ -32,19 +35,35 @@ type Badger interface {
...
+		Badger:       v,
+		hc:           hc,
+		PayloadCache: cache,
+	}, nil
+}
+
+func (s *Datastore) CacheBlock(ctx context.Context, block *structs.CompleteBlockstruct) error {
+	key := structs.PayloadKey{BlockHash: block.Payload.Payload.Data.BlockHash, Slot: structs.Slot(block.Payload.Trace.Message.Slot), Proposer: block.Payload.Trace.Message.ProposerPubkey}
+	s.PayloadCache.Add(key, &block.Payload)
+	s.Logger.With(key).Debug("payload cached")
```
lukanus, line 65: This can be also added to this one log we already write - outside datastore :)

pkg/datastore/datastore.go
```diff
@@ -137,6 +156,13 @@ func (s *Datastore) PutPayload(ctx context.Context, key structs.PayloadKey, payl
 }
 
 func (s *Datastore) GetPayload(ctx context.Context, key structs.PayloadKey) (*structs.BlockBidAndTrace, error) {
+	memPayload, ok := s.PayloadCache.Get(key)
+	if ok {
+		s.Logger.With(key).Debug("payload cache hit")
+		return memPayload, nil
+	}
+	s.Logger.With(key).Debug("payload cache miss")
```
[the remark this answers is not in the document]
aratz-lasa: Good idea

pkg/auction/auctioneer.go
```diff
@@ -0,0 +1,88 @@
...
+
+	auction.mu.Lock()
+	defer auction.mu.Unlock()
+
+	auction.latestBlockByBuilder[block.Payload.Trace.Message.BuilderPubkey] = block
+
+	// always set new value and bigger slot
+	if auction.maxProfit == nil || auction.maxProfit.Header.Trace.Slot < block.Header.Trace.Slot {
+		auction.maxProfit = block
+		a.l.WithField("slot", block.Header.Trace.Slot).WithField("value", block.Header.Trace.Value.String()).Trace("new max bid")
```
lukanus, line 45: Do we really need that at prod?

lukanus (commented):
Looks good @aratz-lasa as it's my PR you need to approve, so feel free to do so :)

pkg/auction/auctioneer.go
aratz-lasa: Mmm good question. This helps to keep track of the auctions, but I think you are right, the logs we have on the block submissions may be enough? In any case, having it as trace log may be useful?

aratz-lasa (approved):
LGTM, it is tested in testnet

PR #59, opened 2022-12-12
GetPayload Performance improvements

lukanus:
What 🕵️‍♀️
Change getPayload so:
 * Verification Signature call would not be enqueued with other verifications, just run synchronously
 * DataAPI PutDelivered write, would be ran asynchronously being outside of the critical path.
Why 🔑
Performance improvements


aratz-lasa (approved):
LGTM, tested on testnet

PR #61, opened 2022-12-14
Streaming layer for high performance bids

aratz-lasa:
# What 🕵️‍♀️
This PR adds a streaming layer to the relay, so that when the relay is distributed it delivers bids and payloads from memory instead of an external DB

# Why 🔑
GetHeader and GetPayload requests are very critical and they have low latency requirements. So, it is important to optimize how fast the bids and payloads are retrieved

# How ⚙️
We add a streaming layer, so that every block a relay receives it streams to the other relays. This way, if the other relays receive a payload request it is very likely they can serve it from memory instead of querying an external DB.

# Testing 🧪


pkg/stream/redis.go
```diff
@@ -0,0 +1,74 @@
...
+}
+
+func (r *RedisPubsub) Publish(ctx context.Context, topic string, data []byte) error {
+	return r.Redis.Publish(ctx, topic, data).Err()
+}
+
+func (r *RedisPubsub) Subscribe(ctx context.Context, topic string) (chan []byte, error) {
+	pubsub := r.Redis.Subscribe(ctx, topic)
+
+	sub := make(chan []byte)
```
lthibault, line 24: Consider making this a `chan string` instead of `chan byte`, and doing the `[]byte` conversion at the call site (in `RunSubscriber`).  There's a good chance this will avoid escaping the stack.

pkg/stream/stream.go
```diff
@@ -0,0 +1,336 @@
...
+func (s *StreamDatastore) RunSubscriber(ctx context.Context) error {
+	blocks, err := s.Pubsub.Subscribe(ctx, s.Config.PubsubTopic)
+	if err != nil {
+		return err
+	}
+
+	sBlock := StreamBlock{}
+	for rawSBlock := range blocks {
+		if err := proto.Unmarshal(rawSBlock, &sBlock); err != nil {
+			s.Logger.Warnf("fail to decode stream block: %s", err.Error())
```
lthibault, line 77: Same as above:  I recommend using `WithError` and avoiding string-formatting in the message.  There's a bunch of these in this PR, so I won't flag every single one :)

pkg/stream/structs.go
```diff
@@ -0,0 +1,100 @@
...
+		Trace: &types.SignedBidTrace{
+			Message: &types.BidTrace{},
+		},
+		Payload: &types.GetPayloadResponse{
+			Data: &types.ExecutionPayload{},
+		},
+	}
+
+	// copy trace
+	copy(block.Trace.Signature[:], streamBlock.Block.Trace.Signature)
```
lthibault, line 67: Is the copy required?  I'm generally in favor of defensive copies on slices, but there's a _lot_ of them, so I start to worry about performance implications.

This seems to be called on each stream element, with O(n) complexity.  Do we have an idea of stream volume?

pkg/structs/structs.go
```diff
@@ -157,12 +157,22 @@ type HeaderAndTrace struct {
 	Trace  *BidTraceWithTimestamp
 }
 
-type BlockBidAndTrace struct {
-	Trace   *types.SignedBidTrace
-	Bid     *types.GetHeaderResponse
+type BlockAndTrace struct {
+	Trace *types.SignedBidTrace
+	//Bid     *types.GetHeaderResponse
```
lthibault, line 162: Remove dead code?

lthibault (changes requested):

cmd/dreamboat/main.go
```diff
@@ -273,20 +330,80 @@ func run() cli.ActionFunc {
...
+			}(streamDs)
+
+			go func(s *stream.StreamDatastore) error {
+				config.Log.With(log.F{
+					"relay-service": "stream-subscriber",
+					"startTimeMs":   time.Since(timeStreamStart).Milliseconds(),
+				}).Info("initialized")
+
+				err := s.RunSubscriber(cContext)
+				if err != nil && !errors.Is(err, context.Canceled) && !errors.Is(err, context.DeadlineExceeded) {
```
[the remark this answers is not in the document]
aratz-lasa: How could **NOT** be both, I'm negating. And I add both, just in case in the future we add timeouts, but I can get rid of the second one, sure

pkg/stream/structs.go
aratz-lasa: The copy is needed because unfortunately one of them is an array, while the other is a slice. Unless there is a way I don't know about

lthibault: Ok, that makes sense.  Unless we're expecting very large data streams, I'm okay to ship this.  Just keep an eye on performance, in case this introduces a sudden severe slowdown.  We might even consider adding a prometheus metric to track this, if it's not too much work.

cmd/dreamboat/main.go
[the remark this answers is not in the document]
lthibault: D'oh 🤦‍♂️ 

Carry on.

aratz-lasa:
Closed for a v2 that is found at #71

PR #63, opened 2022-12-20
Registration cache

lukanus:
# What 🕵️‍♀️
Registration cache for validator registrations

# Why 🔑
To limit the number of registration updates

### DIFF
--- a/pkg/relay/processmanager.go
+++ b/pkg/relay/processmanager.go
@@ -169,7 +169,7 @@ func (rm *ProcessManager) Check(rvg *types.RegisterValidatorRequestMessage) bool
 		return false
 	}
 
-	if uint64(time.Now().Second())-v.Timestamp > uint64(rm.storeTTLHalftimeSeconds+rand.Intn(rm.storeTTLHalftimeSeconds)) {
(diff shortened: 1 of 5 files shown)


pkg/relay/processmanager.go
```diff
@@ -37,11 +41,16 @@ type ProcessManager struct {
 	m ProcessManagerMetrics
 }
 
-func NewProcessManager(l log.Logger, verifySize, storeSize uint) *ProcessManager {
-	rm := &ProcessManager{
-		l:           l,
-		LastRegTime: make(map[string]uint64),
+func NewProcessManager(l log.Logger, verifySize, storeSize uint, registrationCacheSize int) *ProcessManager {
```
aratz-lasa, line 44: Why don't we return `(*ProcessManager, error)`?

aratz-lasa (commented):
I just have one comment, where we should return the struct and the error, instead of returning nil on error.

pkg/relay/processmanager.go
lukanus: In general constructors shouldn't return an error ;) but yeah... you're right, I'll change that

aratz-lasa (approved):
This is an amazing idea @lukanus, amazing work!

aratz-lasa (approved):
Just a small comment to remove a TODO, but LTGMT

PR #64, opened 2022-12-21
Clean logs, add metrics and reorg code

aratz-lasa:
# What 🕵️‍♀️

- Cleans and improves logging
- Adds logs and metrics for understanding when a request fails

# Why 🔑
Currently there are too many logs and there is no clear rules to follow for implementing them. Moreover, there is no understanding on why a GetHeader fails, since it may fail because they query an old slot, or because there was no block submission for the slot

# Testing 🧪


lukanus (approved):
That's a great PR ! :)

pkg/service.go
```diff
@@ -78,33 +78,6 @@ func NewService(l log.Logger, c Config, d Datastore, r Relay, as *AtomicState) *
 	}
 }
 
-// Run creates a relay, datastore and starts the beacon client event loop
-func (s *Service) RunBeacon(ctx context.Context) (err error) {
```
lukanus, line 82: <3

lukanus:
Still good :) I really like the "beacon" migration

pkg/structs/metrics.go
```diff
@@ -0,0 +1,33 @@
...
+	mg[name] = dur
+}
+
+func (mg MetricGroup) ObserveSince(name string, t time.Time) {
+	mg[name] = time.Since(t)
+}
+
+func (mg MetricGroup) Commit(t PrometheusObserver, name string) {
+	for metric, dur := range mg {
+		t.WithLabelValues(name, metric, "").Observe(float64(dur.Milliseconds() / 1_000))
```
lukanus, line 25: that would be wrong. the float second gives you like 0.5 second and on, it's not like 0  for anything below 1

aratz-lasa: _wrong_ -> _not the optimal way_, since the code would work 😉

But yes, thank you, I missed that! Changed 👍

PR #66, opened 2022-12-22
Feature/publish block

aratz-lasa:
# What 🕵️‍♀️
This PR publishes the signed beacon block after a payload delivery

# Why 🔑
This reduces chances of missing payloads due to high-latencies between relay and mev-boost

# Testing 🧪
- [x] test on testnet

### DIFF
--- a/pkg/relay/relay.go
+++ b/pkg/relay/relay.go
@@ -300,8 +300,9 @@ func (rs *Relay) GetPayload(ctx context.Context, payloadRequest *types.SignedBli
 			beaconBlock := structs.SignedBlindedBeaconBlockToBeaconBlock(payloadRequest, payload.Payload.Data)
 			if err := rs.beacon.PublishBlock(beaconBlock); err != nil {
 				logger.WithError(err).Warn("fail to publish block to beacon node")
+			} else {
+				logger.Info("published block to beacon node")
 			}
-			logger.Info("published block to beacon node")
 		}
 
 		if err := rs.d.PutDelivered(ctx, slot, trace, rs.config.TTL); err != nil {
--- a/pkg/relay/relay.go
+++ b/pkg/relay/relay.go
@@ -61,11 +61,17 @@ type RegistrationManager interface {
 	Check(*types.RegisterValidatorRequestMessage) bool
 }
 
+type Beacon interface {
+	PublishBlock(block *types.SignedBeaconBlock) error
+}
+
 type RelayConfig struct {
 	BuilderSigningDomain  types.Domain
 	ProposerSigningDomain types.Domain
 	PubKey                types.PublicKey
 	SecretKey             *bls.SecretKey
+
+	PublishBlock bool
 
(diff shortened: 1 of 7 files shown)


dmarzzz (approved):
LGTM! Can we include a reference to the flashbots relay PR in this PR description to give credit for the idea?

pkg/relay/relay.go
```diff
@@ -287,6 +296,15 @@ func (rs *Relay) GetPayload(ctx context.Context, payloadRequest *types.SignedBli
 
 	// defer put delivered datastore write
 	go func(rs *Relay, slot structs.Slot, trace structs.DeliveredTrace) {
+		if rs.config.PublishBlock {
```
lukanus, line 299: I'm not super sure about that, but if we're doing that - I think this should happen after rs.d.putDelivered - not before

pkg/relay/relay.go
```diff
@@ -287,6 +296,15 @@ func (rs *Relay) GetPayload(ctx context.Context, payloadRequest *types.SignedBli
...
+			beaconBlock := structs.SignedBlindedBeaconBlockToBeaconBlock(payloadRequest, payload.Payload.Data)
+			if err := rs.beacon.PublishBlock(beaconBlock); err != nil {
+				logger.WithError(err).Warn("fail to publish block to beacon node")
+			} else {
+				logger.Info("published block to beacon node")
```
lukanus, line 304: also `rs.l.` :)

pkg/relay/relay.go
aratz-lasa: Why after `PutDelivered`? PutDelivered is only for updating the Data API

pkg/relay/relay.go
aratz-lasa: Why? `logger` is initialized in the beginning of the method, and it contains fields for describing the request

pkg/relay/relay.go
lukanus: Because the integrity of dataAPI imho trumps block publishing, but if you like to keep it like that, not a big deal

aratz-lasa: And does not affect to whether the call will be successful or not. In case of error we simply log it

aratz-lasa: So, you suggest not to publish the block it if the `PutDelivered` call fails? I think the Data API should be relatively decoupled from the relay operations.

lukanus: I was thinking about something more like a precaution, like if publish block (randomly?) panics, to be paranoic about data api having it published.

aratz-lasa: Okay sure, paranoid-drive-development is always good ;) It does not have any drawbacks, so I'll move it to after the call

pkg/relay/relay.go
lukanus: 👍

PR #69, opened 2023-01-09
Fix registerValidator/all metric

aratz-lasa:
# What 🕵️‍♀️
Fixes the observing of the metric `registerValidator-all`

# Why 🔑
Otherwise we don't get real numbers on the time it takes to process registrations

# How ⚙️
Add a `defer`, in order to observe after the function is finished

# Testing 🧪


lukanus (approved):
👍

PR #70, opened 2023-01-11
unwrap error for metrics

aratz-lasa:
# What 🕵️‍♀️
Fix the error label for relay metrics

# Why 🔑
Currently, the errors that relay generates have unique strings, because the error states the exact validator, slot, etc. that caused the error. But, we need a limited and constant set of error messages, otherwise the visualization filtered by errors is not useful.

# How ⚙️
Unwrap the error and unifiy unique messages into higher-level generic messages.

# Testing 🧪


lukanus (approved):
👍

PR #71, opened 2023-01-12
Streaming layer for high performance bids v2

aratz-lasa:
# What 🕵️‍♀️
This PR adds a streaming layer to the relay, so that when the relay is distributed it delivers bids and payloads from memory instead of an external DB

# Why 🔑
GetHeader and GetPayload requests are very critical and they have low latency requirements. So, it is important to optimize how fast the bids and payloads are retrieved

# How ⚙️
We add a streaming layer, so that every block a relay receives it streams to the other relays. This way, if the other relays receive a payload request it is very likely they can serve it from memory instead of querying an external DB.

# Testing 🧪


pkg/stream/redis.go
```diff
@@ -0,0 +1,42 @@
...
+type RedisPubsub struct {
+	Redis  *redis.Client
+	Logger log.Logger
+}
+
+func (r *RedisPubsub) Publish(ctx context.Context, topic string, data []byte) error {
+	return r.Redis.Publish(ctx, topic, data).Err()
+}
+
+func (r *RedisPubsub) Subscribe(ctx context.Context, topic string) (chan []byte, error) {
```
lukanus, line 19: As you never return error here. any chance to consider 
```suggestion
func (r *RedisPubsub) Subscribe(ctx context.Context, topic string, sub chan []byte) {
```
and then just  `go r.Subscribe( .. ) ` somewhere else? 
It may allow then the creation and termination (closing) of the channel outside of this goroutine.  Unless it's essential to do it right here

pkg/stream/stream.go
```diff
@@ -0,0 +1,245 @@
...
+
+func (s *RedisStream) RunPublisherParallel(ctx context.Context, num uint) {
+	for i := uint(0); i < num; i++ {
+		go s.RunPublisher(ctx)
+	}
+}
+
+func (s *RedisStream) RunPublisher(ctx context.Context) error {
+	for {
+		select {
```
lukanus, line 113: I was doing that quite often myself but I think Louis can calculate a proof for you from the statistical perspective it wouldn't matter. and calling multiple selects in one for may have worse performance than go randomised reads.If you like to make it prioritised I would consider just running another goroutine instead. Unless there is some particular data race case that I don't see here?

pkg/stream/stream.go
```diff
@@ -0,0 +1,245 @@
...
+}
+
+func (s *RedisStream) PublishCacheBlock() chan *structs.BlockAndTrace {
+	return s.cacheRequests
+}
+
+func (s *RedisStream) encodeAndPublish(ctx context.Context, block *structs.BlockAndTrace, isCache bool) {
+	timer1 := prometheus.NewTimer(s.m.Timing.WithLabelValues("encodeAndPublish", "encode"))
+	protoBlock := ToStreamBlock(block, isCache, s.Config.ID)
+	rawBlock, err := proto.Marshal(protoBlock)
```
lukanus, line 143:
```suggestion 
	rawBlock, err := proto.Marshal(ToStreamBlock(block, isCache, s.Config.ID))
```
:)

pkg/relay/relay.go
```diff
@@ -377,6 +398,14 @@ func (rs *Relay) SubmitBlock(ctx context.Context, m *structs.MetricGroup, submit
 		return fmt.Errorf("fail to generate contents from block submission: %w", err)
 	}
 
+	if rs.config.Distributed && rs.config.StreamSubmissions {
+		select {
+		case rs.s.PublishStoreBlock() <- &complete.Payload:
```
lukanus, line 403: Should we send a structure here instead of a pointer?

pkg/relay/relay.go
```diff
@@ -144,14 +154,26 @@ func (rs *Relay) GetHeader(ctx context.Context, m *structs.MetricGroup, request
...
 	m.AppendSince(tGet, "getHeader", "get")
 
 	if err := rs.d.CacheBlock(ctx, maxProfitBlock); err != nil {
-		logger.Warnf("fail to cache block: %s", err.Error())
+		logger.WithError(err).Warn("failed to cache block")
 	}
 	logger.Debug("payload cached")
 
+	if rs.config.Distributed { // TODO: run inside goroutine? If so, what context to use?
+		go func() {
```
lukanus, line 165: We're writing to a channel here. isn't this... asynchronous enough ? :)

pkg/datastore/redis.go
```diff
@@ -0,0 +1,47 @@
...
+func payloadKeyToRedisKey(key structs.PayloadKey) string {
+	return fmt.Sprintf("p-%d-%s-%s", key.Slot, key.Proposer, key.BlockHash) // TODO: optimize key size
+}
+
+func decodePayload(data []byte) (*structs.BlockAndTrace, error) {
+	var block structs.BlockAndTrace
+	return &block, json.Unmarshal(data, &block)
+}
+
+func encodePayload(payload *structs.BlockAndTrace) ([]byte, error) {
```
lukanus, line 45: 🤔  I would probably understand a struct matching some interface here. 
This one I don't necessarily get :). If it's a function within the same package, why even bother doing that and not call json.Marshal directly?

pkg/datastore/remotedatastore.go
```diff
@@ -0,0 +1,112 @@
...
+}
+
+func (s *LocalRemoteDatastore) GetPayload(ctx context.Context, key structs.PayloadKey) (*structs.BlockAndTrace, bool, error) {
+	timer0 := prometheus.NewTimer(s.m.Timing.WithLabelValues("getPayload", "all"))
+	defer timer0.ObserveDuration()
+
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	responses := make(chan getPayloadResponse, 2)
```
lukanus, line 54: I think we can Pool that.  Also I don't think I see close on that one :)

pkg/datastore/remotedatastore.go
```diff
@@ -0,0 +1,112 @@
...
+		timer1.ObserveDuration()
+		responses <- getPayloadResponse{block: block, isLocal: false, fromCache: false, err: err}
+	}(ctx, responses)
+
+	for i := 0; i < cap(responses); i++ {
+		var resp getPayloadResponse
+		select {
+		case resp = <-responses:
+		case <-ctx.Done():
+			return &structs.BlockAndTrace{}, false, ctx.Err()
```
lukanus, line 75: That and other returns further down may just leave us with the channel with unread responses. In this particular case even unclosed :)

pkg/stream/stream.go
lthibault: I would recommend shipping as-is, and profiling.  I don't think we're in a pathological case here, because we're not dynamically swapping out channels, and there isn't a channel that is (almost) always in a receivable state.

PR #74, opened 2023-01-17
add processingTime to the slot processing logs

aratz-lasa:
# What 🕵️‍♀️
Add the processing time to the slot processing log

# Why 🔑
We do not have observability on the time it takes to process a slot

# Testing 🧪


dmarzzz (approved):
LGTM

PR #75, opened 2023-01-18
Feature/fast boot

aratz-lasa:
# What 🕵️‍♀️
Adds a flag to speed up the booting up of the relay in exchange for adding temporary inconsistency on the builder_blocks_received endpoint, as well as not loading validator registrations to the cache

# Why 🔑
This avoids slow booting up when the local DB (e.g. badgerdb) is very big

# Testing 🧪


lthibault (commented):
Code looks correct.  Holding off on approval until we sync up on the previous comment.

lthibault (approved):
Addendum, per Slack discussion with @aratz-lasa .  This only affects the *data-api*.

(Mentioning this for posterity)

PR #77, opened 2023-01-25
Fix publishBlock nil values + attach metrics

aratz-lasa:
# What 🕵️‍♀️
This PR fixes 2 things:
- Attaches registration metrics on `main.go`
- Fixes the block struct used in `PublishBlock` call, by replacing nil values with zero values 

# Why 🔑
- If we don't attach metrics on `main.go`, the relay does not publish them to Prometheus
- If we have nil values in `PublishBlock` call, Lighthouse beacon node does not accept the block.


dmarzzz (approved):
LGTM

PR #79, opened 2023-01-27
PublishBlock before storing "payload delivery" in data API

aratz-lasa:
# What 🕵️‍♀️
This PR moves the PublishBlock call to before PutDelivered, in order to speed up as much as possible.

# Why 🔑
If the Storage is congested, then the PublishBlock will take too much time.

### DIFF
--- a/pkg/relay/relay.go
+++ b/pkg/relay/relay.go
@@ -283,7 +283,7 @@ func (rs *Relay) GetPayload(ctx context.Context, m *structs.MetricGroup, payload
 			}
 		}
 
-		if err := rs.d.PutDelivered(ctx, slot, trace, rs.config.TTL); err != nil {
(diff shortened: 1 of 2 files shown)


pkg/relay/relay.go
```diff
@@ -286,6 +282,10 @@ func (rs *Relay) GetPayload(ctx context.Context, m *structs.MetricGroup, payload
 				logger.Info("published block to beacon node")
 			}
 		}
+
+		if err := rs.d.PutDelivered(ctx, slot, trace, rs.config.TTL); err != nil {
```
lukanus, line 286: Just seen it - so, having a parent context this one would be using request's context: 
1) handler running 
2) goroutine spawns 
3) handler finished 
4) Put delivered doesn't really happens, as it would check parent context that is closed already.

lukanus (approved):

lukanus (approved):
👍

PR #80, opened 2023-01-27
Fix/trace timestamp

aratz-lasa:
# What 🕵️‍♀️
Adds Timestamp and TimestampMs to builder block submission data API

# Why 🔑
So that external observers can have greater insights on block submissions

### DIFF
--- a/pkg/relay/relay.go
+++ b/pkg/relay/relay.go
@@ -459,7 +459,8 @@ func (rs *Relay) prepareContents(submitBlockRequest *types.BuilderSubmitBlockReq
 				BlockNumber: s.Payload.Payload.Data.BlockNumber,
 				NumTx:       uint64(len(s.Payload.Payload.Data.Transactions)),
 			},
-			Timestamp: s.Payload.Payload.Data.Timestamp,
+			Timestamp: uint64(time.Now().Second()),
+			TimestampMs: uint64(time.Now().UnixMilli()),
 		},
 	}
 
--- a/pkg/relay/relay.go
+++ b/pkg/relay/relay.go
@@ -459,7 +459,7 @@ func (rs *Relay) prepareContents(submitBlockRequest *types.BuilderSubmitBlockReq
 				BlockNumber: s.Payload.Payload.Data.BlockNumber,
 				NumTx:       uint64(len(s.Payload.Payload.Data.Transactions)),
 			},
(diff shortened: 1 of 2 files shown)


pkg/relay/relay.go
```diff
@@ -459,7 +459,8 @@ func (rs *Relay) prepareContents(submitBlockRequest *types.BuilderSubmitBlockReq
 				BlockNumber: s.Payload.Payload.Data.BlockNumber,
 				NumTx:       uint64(len(s.Payload.Payload.Data.Transactions)),
 			},
-			Timestamp: s.Payload.Payload.Data.Timestamp,
+			Timestamp:   uint64(time.Now().UnixMilli() / 1_000),
```
lthibault, line 462: You've probably already thought of this, but can you tweak the log formatter to get the precision you want?  That's the "cleaner" way.

pkg/relay/relay.go
```diff
@@ -459,7 +459,8 @@ func (rs *Relay) prepareContents(submitBlockRequest *types.BuilderSubmitBlockReq
...
+			TimestampMs: uint64(time.Now().UnixMilli()),
```
lthibault, line 463: Recommend using `time.Duration` (which is what `UnixMilli()` returns, I think), rather than casting as `uint64`.

lthibault (approved):
Approving for expediency, but please consider inline comments.

pkg/relay/relay.go
aratz-lasa: What do you mean by log formatter? This is not related to logging

pkg/relay/relay.go
aratz-lasa: `UnixMilli()` returns `int64`. How are you suggesting exactly to use `time.Duration`?

lthibault: Ah, my bad.  And on second thought, this isn't technically a duration.  Please disregard this comment.  Resolving.

pkg/relay/relay.go
lthibault: Oh, right.  I just assumed it was.  Nevermind then! 🙂

lukanus: Even if it was in logs I would argue that implicit precision changes are "cleaner".

lthibault (approved):

PR #81, opened 2023-01-30
Database Registrations

lukanus:
# What 🕵️‍♀️
Moves registrations into external database (postgres)


pkg/service.go
```diff
@@ -186,51 +196,58 @@ func (s *Service) knownValidatorsUpdateTime() time.Time {
...
-				Entry: &reg,
-			})
-		} else if err != nil && !errors.Is(err, ds.ErrNotFound) {
-			logger.Warn(err)
+		reg, ok := vCache.Get(e.PubKey.PublicKey)
+		if !ok {
+			reg.Entry, err = d.GetRegistration(ctx, e.PubKey.PublicKey)
+			if err != nil {
+				if !errors.Is(err, datastore.ErrNotFound) {
+					logger.Warn(err)
```
aratz-lasa, line 240: Could we add something like. `logger.Warn(fmt.Errorf("fail to register validator %s: %w", pubkey, err))`?

aratz-lasa (approved):
I spent 30-60 minutes reading the code. I tried to be as careful as possible, and lgtm. But we must test it really deep to be sure

pkg/service.go
lukanus: as you wish :)

PR #85, opened 2023-02-03
Clean/api handler

lukanus:
# What 🕵️‍♀️
Cleaned api package and superflous header write error massages

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -114,6 +114,8 @@ func status(w http.ResponseWriter, r *http.Request) {
 
 // proposer related handlers
 func (a *API) registerValidator(w http.ResponseWriter, r *http.Request) {
+	w.Header().Set("Content-Type", "application/json")
+
 	timer := prometheus.NewTimer(a.m.ApiReqTiming.WithLabelValues("registerValidator"))
 	defer timer.ObserveDuration()
 
@@ -156,6 +158,8 @@ func (a *API) registerValidator(w http.ResponseWriter, r *http.Request) {
 }
 
 func (a *API) getHeader(w http.ResponseWriter, r *http.Request) {
+	w.Header().Set("Content-Type", "application/json")
+
 	timer := prometheus.NewTimer(a.m.ApiReqTiming.WithLabelValues("getHeader"))
 	defer timer.ObserveDuration()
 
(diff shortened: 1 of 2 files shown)


pkg/api/api.go
```diff
@@ -183,29 +178,41 @@ func (a *API) getHeader(w http.ResponseWriter, r *http.Request) (int, error) {
 			"slot":     slot,
 			"proposer": proposer,
 		}).WithError(err).Debug("failed getHeader")
-		return http.StatusBadRequest, err
+
+		w.WriteHeader(http.StatusBadRequest)
```
aratz-lasa, line 182: can we extract this code into a function? In order to avoid code duplication

pkg/api/api.go
```diff
@@ -183,29 +178,41 @@ func (a *API) getHeader(w http.ResponseWriter, r *http.Request) (int, error) {
...
+		})
+		return
 	}
 
 	m.Observe(a.m.RelayTiming)
 
 	if err = json.NewEncoder(w).Encode(response); err != nil {
 		a.l.WithError(err).WithField("path", r.URL.Path).Debug("failed to write response")
 		a.m.ApiReqCounter.WithLabelValues("getHeader", "500", "response encode").Inc()
-		return http.StatusInternalServerError, err
```
aratz-lasa, line 194: Why did you delete this?

pkg/api/api.go
```diff
@@ -183,29 +178,41 @@ func (a *API) getHeader(w http.ResponseWriter, r *http.Request) (int, error) {
...
+		return
 	}
 
 	a.m.ApiReqCounter.WithLabelValues("getHeader", "200", "").Inc()
-	return http.StatusOK, nil
```
aratz-lasa, line 198: Why did you delete this?

aratz-lasa (changes requested):

pkg/api/api.go
lukanus: as earlier - those functions are just plain http handlers right now.

pkg/api/api.go
lukanus: If you like. I'm not sure it's needed in this particular case, but If you prefer - I'm fine.

aratz-lasa (approved):

PR #87, opened 2023-02-06
Builder submission ratelimit

lukanus:
# What 🕵️‍♀️
This work adds a rate limit in preparation for adding support for external builders.

# Testing 🧪

### DIFF
--- a/cmd/dreamboat/main.go
+++ b/cmd/dreamboat/main.go
@@ -7,6 +7,7 @@ import (
 	"net/http"
 	"os"
 	"os/signal"
+	"strings"
 	"syscall"
 
 	"time"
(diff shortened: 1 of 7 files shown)


cmd/dreamboat/main.go
```diff
@@ -200,6 +201,24 @@ var flags = []cli.Flag{
...
 		EnvVars: []string{"RELAY_FAST_BOOT"},
 	},
+	&cli.StringFlag{
+		Name:    "relay-allow-listed-builder",
+		Usage:   "comma separated list of allowed builder pubkeys",
+		Value:   "",
+		EnvVars: []string{"RELAY_ALLOW_LISTED_BUILDER"},
+	},
+	&cli.DurationFlag{
+		Name:    "relay-submission-limit-rate",
```
dmarzzz, line 211: I think X/s for regular rate and Y burst makes more sense. The default value here should also be 2blocks/second as that is industry standard as seen from the Flashbots relay.

dmarzzz (commented):
LGTM w/ 1 inline comment and also can we add a section to the readme: "External Builder Support" which explains how to set these commands, only 1 or 2 sentences should do! Also great call on doing the rate limiting before introspecting any of the block payload.

Also for posterity on the term allow-list builder, Blocknative hopes to move dreamboat away from being a relay that enables special treatment of a single builder but internally this was chosen as the best path forward until less trusted support of external builders can be developed.

dmarzzz (approved):

PR #89, opened 2023-02-09
Add log to record invalid parenthash

aratz-lasa:
# What 🕵️‍♀️
Adds a log statement to record whenever an invalid parenthash is requested on GetHeader

# Why 🔑
This way we can debug if we fail to deliver bids due to invalid parenthash

### DIFF
--- a/pkg/relay/relay.go
+++ b/pkg/relay/relay.go
@@ -186,7 +186,9 @@ func (rs *Relay) GetHeader(ctx context.Context, m *structs.MetricGroup, request 
 	header := maxProfitBlock.Header
 
 	if header.Header == nil || (header.Header.ParentHash != parentHash) {
-		logger.Debug(ErrBadHeader.Error())
+		if header.Header.ParentHash != parentHash {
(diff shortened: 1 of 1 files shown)


pkg/relay/relay.go
```diff
@@ -186,7 +186,9 @@ func (rs *Relay) GetHeader(ctx context.Context, m *structs.MetricGroup, request
 	header := maxProfitBlock.Header
 
 	if header.Header == nil || (header.Header.ParentHash != parentHash) {
-		logger.Debug(ErrBadHeader.Error())
+		if header.Header.ParentHash != parentHash {
+			logger.WithField("expected", parentHash).WithField("got", parentHash).Debug("invalid parentHash")
```
lukanus, line 190: I wonder shouldn't we just add this one for every log above.

aratz-lasa: Not really, because we don't have `header.Header.ParentHash`, until we get it from the DB

lukanus (approved):
👍

PR #92, opened 2023-02-21
Add stream layer to main - V3

aratz-lasa:
# What 🕵️‍♀️

# Why 🔑

# How ⚙️

# Testing 🧪


pkg/relay/relay.go
```diff
@@ -216,6 +226,18 @@ func (rs *Relay) GetHeader(ctx context.Context, m *structs.MetricGroup, request
 		return nil, ErrInternal
 	}
 
+	if rs.config.Distributed { // TODO: run inside goroutine? If so, what context to use?
+		go func() {
```
aratz-lasa, line 230: This is the endpoint with the highest performance requirements. The entire HTTP RTT must not exceed 900ms. So either we run it as a goroutine or set a default: skip the streaming. This is the best option. However, I'm open to talking with you about it.

pkg/stream/stream.go
```diff
@@ -0,0 +1,241 @@
...
+
+	timer2 := prometheus.NewTimer(s.m.Timing.WithLabelValues("encodeAndPublish", "publish"))
+	defer timer2.ObserveDuration()
+	if err := s.Pubsub.Publish(ctx, s.Config.PubsubTopic, rawBlock); err != nil {
+		s.Logger.Warnf("fail to encode encode and stream block: %s", err.Error())
+		return
+	}
+}
+
+func (s *RedisStream) cachePayload(ctx context.Context, ds Datastore, payload *structs.BlockAndTrace) error {
```
lukanus, line 155: How do you feel about moving this cache somewhere else? 
So it would become separate entity that is unrelated to both datastore and stream.

pkg/stream/structs.go
```diff
@@ -0,0 +1,100 @@
...
+	"github.com/flashbots/go-boost-utils/types"
+)
+
+func ToStreamBlock(block *structs.BlockAndTrace, isCache bool, source string) *StreamBlock {
+	var transactions [][]byte
+	for _, tx := range block.Payload.Data.Transactions {
+		transactions = append(transactions, tx)
+	}
+
+	return &StreamBlock{
```
lukanus, line 15: 👍

pkg/datastore/lrdatastore.go
```diff
@@ -0,0 +1,159 @@
...
+
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	responses := chanPool.Get().(chan getPayloadResponse)
+	defer chanPool.Put(responses)
+
+	id := uuid.NewString()
+
+	go func(ctx context.Context, resp chan getPayloadResponse, id string) {
```
lukanus, line 92: If you change the above to be the same interface this function becomes possible to just call twice without copypasta

pkg/datastore/lrdatastore.go
```diff
@@ -0,0 +1,159 @@
...
+
+	return nil, false, fmt.Errorf("payload not found")
+}
+
+func (s *LocalRemoteDatastore) PutPayload(ctx context.Context, key structs.PayloadKey, payload *structs.BlockAndTrace, ttl time.Duration) error {
+	timer0 := prometheus.NewTimer(s.m.Timing.WithLabelValues("putPayload", "all"))
+	defer timer0.ObserveDuration()
+
+	timer1 := prometheus.NewTimer(s.m.Timing.WithLabelValues("putPayload", "remoteStore"))
+	if err := s.Remote.PutPayload(ctx, key, payload, ttl); err != nil {
```
lukanus, line 150: shouldn't we guarantee that both would be stored "no matter what" ? So if remote fails we still store regardless??

pkg/datastore/lrdatastore.go
```diff
@@ -0,0 +1,159 @@
...
+		timer1 := prometheus.NewTimer(s.m.Timing.WithLabelValues("getPayload", "remote"))
+		block, err := s.Remote.GetPayload(ctx, key)
+		timer1.ObserveDuration()
+		select {
+		case responses <- getPayloadResponse{id: id, block: block, isLocal: false, fromCache: false, err: err}:
+		case <-ctx.Done():
+		}
+	}(ctx, responses, id)
+
+	for i := 0; i < 2; i++ {
```
lukanus, line 112: I'm not sure you need any number here, just for simple for loop would be probably enough, as you would like to terminate it either on any successful compatible(id) return or ctx.Done() . You wouldn't need a counter as well

pkg/stream/stream.go
aratz-lasa: Why would you like to do that?

pkg/datastore/lrdatastore.go
```diff
@@ -0,0 +1,159 @@
...
+
+	"github.com/flashbots/go-boost-utils/types"
+	"github.com/google/uuid"
+	"github.com/lthibault/log"
+	"github.com/prometheus/client_golang/prometheus"
+
+	"github.com/blocknative/dreamboat/pkg/structs"
+)
+
+type LocalDatastore interface {
```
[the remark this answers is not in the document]
aratz-lasa: We cannot go with the second option, because we need to differentiate local and remote explicitly—this way we know that we need to query local before remote, etc.

pkg/stream/stream.go
```diff
@@ -0,0 +1,275 @@
...
+			Timestamp: payload.Payload.Data.Timestamp,
+		},
+	}
+
+	b, err := json.Marshal(h)
+	if err != nil {
+		return err
+	}
+
+	return ds.PutHeader(ctx, structs.HeaderData{
```
lukanus, line 262: Why do we even need headers anymore if we're storing whole payloads?

pkg/relay/relay.go
```diff
@@ -145,6 +164,26 @@ func NewRelay(l log.Logger, config RelayConfig, beacon Beacon, cache ValidatorCa
...
+}
+
+func (rs *Relay) updateSlotDelivered(slot structs.Slot) {
+	rs.slotMu.Lock()
+	defer rs.slotMu.Unlock()
+
+	if rs.slotDelivered < slot {
+		rs.slotDelivered = slot
+	}
+}
```
lukanus, line 185: `slotDelivered` is just an integer I don't thing we need to lock here. Especially that  `updateSlotDelivered()` is triggered by select so is a synchronous operation. 
I think  this should be enough:
```go 

select {
		case slot := <-rs.s.SlotDeliveredChan():
                      if rs.slotDelivered < slot {
        		rs.slotDelivered = slot
	             }
		case <-ctx.Done():
			return ctx.Err()
		}
```

pkg/relay/submit.go
```diff
@@ -75,6 +75,12 @@ func (rs *Relay) SubmitBlock(ctx context.Context, m *structs.MetricGroup, submit
 }
 
 func (rs *Relay) isPayloadDelivered(ctx context.Context, slot uint64) (err error) {
+	// check if delivered in stream layer
+	if rs.isPayloadDeliveredInStream(slot) {
```
lukanus, line 79: as above - I dont think we need anything more than just `slot <= uint64(rs.slotDelivered)` here

pkg/relay/submit.go
```diff
@@ -190,6 +202,14 @@ func (rs *Relay) storeSubmission(ctx context.Context, m *structs.MetricGroup, su
 	}
 	m.AppendSince(tPutHeader, "submitBlock", "putHeader")
 
+	if rs.config.Distributed && rs.config.StreamSubmissions {
+		select {
+		case rs.s.PublishStoreBlock() <- complete.Payload:
```
lukanus, line 207: wouldn't  `rs.s.PublishStoreBlock(ctx, complete.Payload)` 
and then 
```go
func (s *Client) PublishStoreBlock(ctx context.Context(), block something.Payload) error {
        select {
		case  s.storeRequests <-  block:
		case <-ctx.Done():
			return ctx.Err()
		}
}
```
be more readable?

cmd/dreamboat/main.go
```diff
@@ -373,6 +449,51 @@ func run() cli.ActionFunc {
...
 		}
 
+		if c.Bool("relay-distribution") {
+			timeStreamStart := time.Now()
+			redisClient := redis.NewClient(&redis.Options{
+				Addr: c.String("relay-distribution-redis-uri"),
+			})
+
+			// init datastore
+			bds = &blockRedis.RedisDatastore{Redis: redisClient}
```
lukanus, line 459: I think this one should be external  under 
```go
if c.String("relay-distribution-redis-uri") != "" {
    redisClient := redis.NewClient(&redis.Options{
        Addr: c.String("relay-distribution-redis-uri"),
    })

   // init datastore
   ds = &blockRedis.RedisDatastore{Redis: redisClient}
} else {
    ... 
    ds, err = datastore.NewDatastore(&datastore.TTLDatastoreBatcher{TTLDatastore: 
    storage}, storage.DB, hc, c.Int("relay-payload-cache-size"))
}
```

PR #93, opened 2023-02-24
add headers preload optimization

lukanus:
# What 🕵️‍♀️
Optimizing header preload, by only fetching last 5 heights

### DIFF
--- a/pkg/datastore/headers.go
+++ b/pkg/datastore/headers.go
@@ -367,7 +367,7 @@ func (s *Datastore) getOrphanedHeaders(ctx context.Context, slot uint64) (items 
 func (s *Datastore) FixOrphanHeaders(ctx context.Context, latestSlot uint64, ttl time.Duration) error {
 	buff := new(bytes.Buffer)
 	for slot := latestSlot - FixSlotLag; slot < latestSlot+1; slot++ {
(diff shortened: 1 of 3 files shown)


pkg/datastore/headers.go
```diff
@@ -338,110 +339,78 @@ type LoadItem struct {
...
-				if i > 0 {
-					buff.WriteString(",")
-				}
-				io.Copy(buff, bytes.NewReader(payload.Content))
-				hnt := structs.HeaderAndTrace{}
-				if err := json.Unmarshal(payload.Content, &hnt); err != nil {
-					return err
-				}
+	for slot := latestSlot - FixSlotLag; slot < latestSlot+1; slot++ {
+		items, err := s.getOrphanedHeaders(ctx, latestSlot)
```
aratz-lasa, line 370: shouldn't be `s.getOrphanedHeaders(ctx, slot)`?

aratz-lasa (changes requested):

pkg/datastore/headers.go
lukanus: Ah yeah ... you got me

aratz-lasa (approved):

PR #94, opened 2023-02-28
Extract beacon into package

aratz-lasa:
# What 🕵️‍♀️
This PR extracts the beacon logic into its own package under pkg/beacon

# Why 🔑
We can extract the beacon logic into its own component and reutilize it in a more efficient way

# Testing 🧪


pkg/beacon/client/multi.go
```diff
@@ -0,0 +1,202 @@
...
+
+	"github.com/blocknative/dreamboat/metrics"
+	"github.com/blocknative/dreamboat/pkg/structs"
+	"github.com/flashbots/go-boost-utils/types"
+	"github.com/lthibault/log"
+	uberatomic "go.uber.org/atomic"
+)
+
+var (
+	_                    BeaconNode = (*beaconClient)(nil) // type constraint
```
lukanus, line 17: I don't think we need this :)

pkg/beacon/client/multi.go
```diff
@@ -0,0 +1,202 @@
...
+		}
+
+		return genesisInfo, nil
+	}
+
+	return genesisInfo, err
+}
+
+func (b *MultiBeaconClient) Endpoint() string {
+	return b.clientsByLastResponse()[0].Endpoint()
```
lukanus, line 165: So. Under some degree of misuse of this structure you'll get nilpointer here. 
because clientsByLastResponse would return b.Clients that can be nil. It's better safe than sorry :)

pkg/beacon/client/multi.go
```diff
@@ -0,0 +1,202 @@
...
+			continue
+		}
+
+		return nil
+	}
+
+	return err
+}
+
+func (b *MultiBeaconClient) AttachMetrics(m *metrics.Metrics) {
```
lukanus, line 198: It would be better for Clients to take care of it themselves wherever they are initialized. We don't need this in an interface as well.

pkg/beacon/manager.go
```diff
@@ -31,44 +31,42 @@ type ValidatorCache interface {
...
+	GetProposerDuties(structs.Epoch) (*bcli.RegisteredProposersResponse, error)
+	SyncStatus() (*bcli.SyncStatusPayloadData, error)
+	KnownValidators(structs.Slot) (bcli.AllValidatorsResponse, error)
+	Genesis() (structs.GenesisInfo, error)
+	PublishBlock(block *types.SignedBeaconBlock) error
+}
 
-	once  sync.Once
-	ready chan struct{}
+type State interface {
```
lukanus, line 43: 👍

pkg/beacon/manager.go
```diff
@@ -31,44 +31,42 @@ type ValidatorCache interface {
...
+type Manager struct {
+	Log log.Logger
 }
 
-func (s *Service) setReady() {
-	select {
-	case <-s.Ready():
-	default:
-		close(s.ready)
+func NewManager(l log.Logger, as *AtomicState) *Manager {
```
lukanus, line 63:
```suggestion
func NewManager(l log.Logger) *Manager {
```

cmd/dreamboat/main.go
```diff
@@ -459,9 +461,9 @@ func run() cli.ActionFunc {
 		}).Info("initialized")
 
 		cContext, cancel := context.WithCancel(c.Context)
-		go func(s *pkg.Service) error {
+		go func(s *beacon.Manager) error {
```
lukanus, line 464: You've added wait to the state structure, why would we need this one any more?

pkg/relay/relay.go
```diff
@@ -52,7 +52,9 @@ type ValidatorCache interface {
 }
 
 type State interface {
-	Beacon() *structs.BeaconState
+	KnownValidators() structs.ValidatorsState
```
lukanus, line 55: <3

lukanus (approved):

PR #97, opened 2023-03-02
Feature/randao

aratz-lasa:
# What 🕵️‍♀️
Implements randao validation on block submissions

# Why 🔑
Otherwise, it may cause security vulnerabilities

# How ⚙️
Query the randao from the beacon every slot

# Testing 🧪


pkg/beacon/client/client.go
```diff
@@ -137,6 +137,18 @@ func (b *beaconClient) Genesis() (structs.GenesisInfo, error) {
 	return resp.Data, err
 }
 
+func (b *beaconClient) Randao(slot structs.Slot) (string, error) {
+	resp := new(GetRandaoResponse)
+	u := *b.beaconEndpoint
```
lukanus, line 142: wouldn't it be easier to store url.URL instead of pointer, than dereferencing this in every function? Asking for a friend....

lukanus (approved):
:+1

PR #99, opened 2023-03-07
Disable simulation if configuration is not set

lukanus:
# What 🕵️‍♀️ 
Do not perform any block simulation if simulation endpoints are not set


aratz-lasa (approved):
LGTM

PR #102, opened 2023-03-09
Shapella/fork version

aratz-lasa:
# What 🕵️‍♀️

# Why 🔑

# How ⚙️

# Testing 🧪

### DIFF
--- a/pkg/structs/structs.go
+++ b/pkg/structs/structs.go
@@ -272,7 +272,23 @@ const (
 	ForkAltair ForkVersion = iota
 	ForkBellatrix
 	ForkCapella
+	ForkUnknown
 )
+
+func (v ForkVersion) String() string {
+	switch v {
+	case ForkAltair:
+		return "altair"
+	case ForkBellatrix:
+		return "bellatrix"
+	case ForkCapella:
+		return "capella"
+	case ForkUnknown:
+		return "unknown"
+	default:
+		return ""
+	}
+}
 
 func (fs ForkState) IsCapella(slot Slot) bool {
 	return slot.Epoch() >= fs.CapellaEpoch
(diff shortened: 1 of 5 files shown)


pkg/structs/structs.go
```diff
@@ -272,27 +272,45 @@ const (
 	ForkAltair ForkVersion = iota
 	ForkBellatrix
 	ForkCapella
+	ForkUnknown
```
lukanus, line 275: I wouldn't set Unknown as if you like we can make it 0 elem

pkg/structs/structs.go
```diff
@@ -272,27 +272,45 @@ const (
...
+	case ForkAltair:
+		return "altair"
+	case ForkBellatrix:
+		return "bellatrix"
+	case ForkCapella:
+		return "capella"
+	case ForkUnknown:
+		return "unknown"
+	default:
+		return ""
```
lukanus, line 289:
```suggestion 
	default:
		return "unknown"
```

pkg/structs/structs.go
```diff
@@ -272,27 +272,45 @@ const (
...
-	case epoch < uint64(fs.BellatrixEpoch):
+func (fs ForkState) Version(slot Slot) ForkVersion {
+	if fs.IsAltair(slot) {
 		return ForkAltair
+	} else if fs.IsBellatrix(slot) {
+		return ForkBellatrix
+	} else if fs.IsCapella(slot) {
+		return ForkCapella
+	} else {
+		return ForkUnknown
```
lukanus, line 313:
```suggestion
 } 
 return ForkUnknown
```

pkg/structs/structs.go
```diff
@@ -272,27 +272,45 @@ const (
...
+}
 
-	switch {
-	case epoch >= uint64(fs.CapellaEpoch):
-		return ForkCapella
-	case epoch >= uint64(fs.BellatrixEpoch) && epoch < uint64(fs.CapellaEpoch):
-		return ForkBellatrix
-	case epoch < uint64(fs.BellatrixEpoch):
+func (fs ForkState) Version(slot Slot) ForkVersion {
+	if fs.IsAltair(slot) {
```
lukanus, line 306: I know it would be nitpick but - I think we will most likely see newest blocks here, so the if should be in reverse order :)

PR #103, opened 2023-03-10
Add proposer pubkey check in getheader

aratz-lasa:
# What 🕵️‍♀️
This PR adds a check on GetHeader, so that it checks whether the proposer pubkey is the expeceted one

# Why 🔑
Otherwise it does not return the consequent payload

# Testing 🧪


lukanus (approved):
👍

PR #106, opened 2023-03-15
Add intermediary structure for withdraws encoding

lukanus:
# What 🕵️‍♀️ 
Add intermediary structure for withdraws encoding


lukanus:
Although this produces the same sum - lets keep this for consistency

PR #107, opened 2023-03-17
Shapella/fix dataapi

aratz-lasa:
# What 🕵️‍♀️

# Why 🔑

# How ⚙️

# Testing 🧪


lukanus:
That would be invalid after last PR, as we're passing different structure to datastore, so there is no need to change HeaderAndTrace onto interface

PR #114, opened 2023-03-23
Add logic to run simulation validation asynchronously

aratz-lasa:
# What 🕵️‍♀️
This PR validates blocks asynchronously

# Why 🔑
This is for improving performance on block submissions

### DIFF
--- a/relay/submit.go
+++ b/relay/submit.go
@@ -58,6 +58,18 @@ func (rs *Relay) SubmitBlock(ctx context.Context, m *structs.MetricGroup, sbr st
 	}
 	m.AppendSince(tCheckRegistration, "submitBlock", "checkRegistration")
 
+	valErr := make(chan error, 1)
+	go func(ctx context.Context, gasLimit uint64, sbr structs.SubmitBlockRequest, chErr chan error) {
+		defer close(chErr)
+
+		tValidateBlock := time.Now()
+		if err := rs.validateBlock(ctx, gasLimit, sbr); err != nil {
+			chErr <- err
+			return
+		}
+		m.AppendSince(tValidateBlock, "submitBlock", "validateBlock")
+	}(ctx, gasLimit, sbr, valErr)
+
 	tVerify := time.Now()
 	if err := rs.verifySignature(ctx, sbr); err != nil {
 		return err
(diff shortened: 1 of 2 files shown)


relay/submit.go
```diff
@@ -70,11 +82,10 @@ func (rs *Relay) SubmitBlock(ctx context.Context, m *structs.MetricGroup, sbr st
 		return fmt.Errorf("failed to verify withdrawals: %w", err)
 	}
 
-	tValidateBlock := time.Now()
-	if err := rs.validateBlock(ctx, gasLimit, sbr); err != nil {
+	// wait for validations
+	if err := <-valErr; err != nil {
```
lukanus, line 86: select on error and context Done()? maybe?

relay/submit.go
```diff
@@ -58,6 +58,18 @@ func (rs *Relay) SubmitBlock(ctx context.Context, m *structs.MetricGroup, sbr st
...
+	valErr := make(chan error, 1)
+	go func(ctx context.Context, gasLimit uint64, sbr structs.SubmitBlockRequest, chErr chan error) {
+		defer close(chErr)
+
+		tValidateBlock := time.Now()
+		if err := rs.validateBlock(ctx, gasLimit, sbr); err != nil {
+			chErr <- err
+			return
+		}
+		m.AppendSince(tValidateBlock, "submitBlock", "validateBlock")
```
lukanus, line 70: I wonder would this work properly with ObserveWithError. Like as it's async, shouldn't we call observe on that directly

relay/submit.go
```diff
@@ -58,6 +58,18 @@ func (rs *Relay) SubmitBlock(ctx context.Context, m *structs.MetricGroup, sbr st
 	}
 	m.AppendSince(tCheckRegistration, "submitBlock", "checkRegistration")
 
+	valErr := make(chan error, 1)
+	go func(ctx context.Context, gasLimit uint64, sbr structs.SubmitBlockRequest, chErr chan error) {
+		defer close(chErr)
```
lukanus, line 63: 👍

relay/submit.go
aratz-lasa: I added a lock to `MetricGroup`

PR #117, opened 2023-03-27
Submit old slot

aratz-lasa:
# What 🕵️‍♀️
This PR allows submitting blocks from old slots

# Testing 🧪

### DIFF
--- a/beacon/manager.go
+++ b/beacon/manager.go
@@ -60,7 +60,7 @@ type State interface {
 	SetWithdrawals(uint64, structs.WithdrawalsState)
 
 	SetRandao(structs.RandaoState)
-	Randao(uint64) string
+	Randao(uint64) structs.RandaoState
 
 	Fork() structs.ForkState
 	SetFork(structs.ForkState)
@@ -207,7 +207,7 @@ func (s *Manager) Run(ctx context.Context, state State, client BeaconClient, d D
 				"numDuties":                 len(duties.ProposerDutiesResponse),
 				"numKnownValidators":        len(validators.KnownValidators),
 				"knownValidatorsUpdateTime": state.KnownValidatorsUpdateTime(),
-				"randao":                    state.Randao(uint64(headSlot)),
+				"randao":                    state.Randao(uint64(headSlot)).Randao,
 				"processingTimeMs":          time.Since(t).Milliseconds(),
 				"withdrawalsRoot":           state.Withdrawals(uint64(headSlot)).Root.String(),
 				"fork":                      state.Fork().Version(headSlot).String(),
--- a/beacon/manager.go
+++ b/beacon/manager.go
@@ -59,7 +59,7 @@ type State interface {
 	Withdrawals(uint64) structs.WithdrawalsState
 	SetWithdrawals(uint64, structs.WithdrawalsState)
 
-	SetRandao(uint64, string)
+	SetRandao(structs.RandaoState)
 	Randao(uint64) string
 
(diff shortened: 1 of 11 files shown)


relay/submit.go
```diff
@@ -245,13 +242,13 @@ func verifyBlock(sbr structs.SubmitBlockRequest, beaconState State) (retry bool,
 		return false, fmt.Errorf("%w: got %d, expected %d", ErrInvalidTimestamp, sbr.Timestamp(), expectedTimestamp)
```
lukanus, line 242: I think this may also fail if we check for exact comparison

aratz-lasa: Not really, because timestamp is only calculated based on Genesis + the slot of `submitBlockRequest`

beacon/manager.go
```diff
@@ -56,11 +56,11 @@ type State interface {
...
 	SetHeadSlot(structs.Slot)
 
-	Withdrawals() structs.WithdrawalsState
-	SetWithdrawals(structs.WithdrawalsState)
+	Withdrawals(uint64) structs.WithdrawalsState
+	SetWithdrawals(structs.WithdrawalsState, uint64)
 
-	SetRandao(string)
-	Randao() string
+	SetRandao(string, uint64)
```
lukanus, line 62: nitpick - (slot, value) rather than (value, slot) . As slot is more like a key, or context you're passing.

validators/validators.go
```diff
@@ -20,8 +21,12 @@ var (
 	ErrInvalidTimestamp = errors.New("invalid timestamp")
 )
 
+type MultiSlotState interface {
+	Get(uint64) State
+}
+
 type State interface {
-	Duties() structs.DutiesState
+	Duties(uint64) structs.DutiesState
```
lukanus, line 29:
```suggestion
	Duties(uint64) structs.DutiesState
	HeadSlot() uint64
```
? :)

relay/submit.go
```diff
@@ -266,13 +265,13 @@ func verifyWithdrawals(state State, submitBlockRequest structs.SubmitBlockReques
 		return types.Root{}, false, nil
 	}
 
-	withdrawalState := state.Withdrawals()
+	withdrawalState := state.Withdrawals(submitBlockRequest.Slot() - 1)
```
lukanus, line 268: - 1  here and we keep the +1 in the line below?

relay/submit.go
```diff
@@ -36,16 +37,14 @@ func (rs *Relay) SubmitBlock(ctx context.Context, m *structs.MetricGroup, sbr st
...
 		"blockHash":      sbr.BlockHash(),
+		"headSlot":       rs.beaconState.HeadSlot(),
 		"slot":           sbr.Slot(),
+		"slotDiff":       int64(sbr.Slot()) - int64(rs.beaconState.HeadSlot()),
 		"proposer":       sbr.ProposerPubkey(),
 		"bid":            value.String(),
 		"withdrawalsNum": len(sbr.Withdrawals()),
 	})
 
-	if rs.lastDeliveredSlot.Load() >= sbr.Slot() {
```
lukanus, line 45: 👍

lukanus (approved):

validators/validators.go
aratz-lasa: I tried to do this, but it becomes an infinite rabbit hole, so I will open a consequent PR for getting rid of `structs.Slot` and `structs.Epoch` entirely

PR #118, opened 2023-03-29
Add additional checks

aratz-lasa:
# What 🕵️‍♀️
This PR adds additional checks to block submissions, and to GetHeader and GetPayload requests.

# Why 🔑
It is necessary to reinforce the checks to ensure the bids and payloads being served are valid.

### DIFF
--- a/structs/blocks.go
+++ b/structs/blocks.go
@@ -7,11 +7,11 @@ import (
 )
 
 type SubmitBlockRequest interface {
-	BidTrace() types.BidTrace
-
 	Slot() uint64
 	BlockHash() types.Hash
 	ParentHash() types.Hash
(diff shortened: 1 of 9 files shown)


structs/blocks.go
```diff
@@ -7,8 +7,11 @@ import (
 )
 
 type SubmitBlockRequest interface {
+	BidTrace() types.BidTrace
```
lukanus, line 10: I was trying hard not to do that, and limit this to a per usage getter, can we create TraceBlockHash and TraceParentHash by any chance ?:) Or create those validates inside `SubmitBlockRequest` ? Like the comparison function that would do the checks internally ? :)

aratz-lasa: Sure

PR #126, opened 2023-04-04
Warehouse

aratz-lasa:
# What 🕵️‍♀️
This PR implements a service for exporting data to the filesystem. Specifically exports the payloads of every block submission. 

# Why 🔑
This is an option, in case users want to inspect and digest data related to block submissions

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -247,7 +247,7 @@ func (a *API) getPayload(w http.ResponseWriter, r *http.Request) {
 		}
 	case structs.ForkBellatrix:
 		var breq bellatrix.SignedBlindedBeaconBlock
-		if err := json.NewDecoder(r.Body).Decode(&breq); err != nil {
+		if err := json.NewDecoder(bytes.NewReader(b)).Decode(&breq); err != nil {
 			a.m.ApiReqCounter.WithLabelValues("getPayload", "400", "payload decode").Inc()
 			writeError(w, http.StatusBadRequest, errors.New("invalid getPayload request bellatrix decode"))
 			return
@@ -340,7 +340,7 @@ func (a *API) submitBlock(w http.ResponseWriter, r *http.Request) {
 		}
 	case structs.ForkBellatrix:
 		var breq bellatrix.SubmitBlockRequest
-		if err := json.NewDecoder(r.Body).Decode(&breq); err != nil {
+		if err := json.NewDecoder(bytes.NewReader(b)).Decode(&breq); err != nil {
 			a.m.ApiReqCounter.WithLabelValues("submitBlock", "400", "payload decode").Inc()
 			writeError(w, http.StatusBadRequest, errors.New("invalid submitblock request bellatrix decode"))
 			return
--- a/api/api.go
+++ b/api/api.go
@@ -726,6 +726,8 @@ func unwrapError(err error, defaultMsg string) error {
 		return relay.ErrInvalidRandao
 	} else if errors.Is(err, relay.ErrLateRequest) {
 		return relay.ErrLateRequest
+	} else if errors.Is(err, relay.ErrInvalidExecutionPayload) {
+		return relay.ErrInvalidExecutionPayload
 	} else if errors.Is(err, validators.ErrInvalidSignature) {
 		return validators.ErrInvalidSignature
 	} else if errors.Is(err, validators.ErrUnknownValidator) {
--- a/api/api.go
+++ b/api/api.go
@@ -11,6 +11,7 @@ import (
 	"io/ioutil"
 	"net/http"
 	"strconv"
+	"strings"
 
 	"github.com/flashbots/go-boost-utils/types"
 	"github.com/gorilla/mux"
@@ -43,7 +44,8 @@ const (
 )
 
 const (
-	DataLimit = 450
+	DataLimit       = 450
+	ErrorsOnDisable = false
 )
 
 var (
@@ -52,11 +54,11 @@ var (
 
 type Relay interface {
 	// Proposer APIs (builder spec https://github.com/ethereum/builder-specs)
-	GetHeader(context.Context, *structs.MetricGroup, structs.HeaderRequest) (structs.GetHeaderResponse, error)
-	GetPayload(context.Context, *structs.MetricGroup, structs.SignedBlindedBeaconBlock) (structs.GetPayloadResponse, error)
+	GetHeader(context.Context, *structs.MetricGroup, structs.UserContent, structs.HeaderRequest) (structs.GetHeaderResponse, error)
+	GetPayload(context.Context, *structs.MetricGroup, structs.UserContent, structs.SignedBlindedBeaconBlock) (structs.GetPayloadResponse, error)
 
 	// Builder APIs (relay spec https://flashbots.notion.site/Relay-API-Spec-5fb0819366954962bc02e81cb33840f5)
-	SubmitBlock(context.Context, *structs.MetricGroup, structs.SubmitBlockRequest) error
+	SubmitBlock(context.Context, *structs.MetricGroup, structs.UserContent, structs.SubmitBlockRequest) error
 
 	// Data APIs
 	GetPayloadDelivered(context.Context, structs.PayloadTraceQuery) ([]structs.BidTraceExtended, error)
@@ -90,16 +92,19 @@ type API struct {
 	lim RateLimitter
 
 	m *APIMetrics
+
+	enabled *EnabledEndpoints
 }
 
-func NewApi(l log.Logger, r Relay, reg Registrations, st State, lim RateLimitter) (a *API) {
+func NewApi(l log.Logger, enabled *EnabledEndpoints, r Relay, reg Registrations, st State, lim RateLimitter) (a *API) {
 	a = &API{
-		l:   l,
-		r:   r,
-		reg: reg,
-		st:  st,
-		lim: lim,
-		m:   &APIMetrics{}}
+		l:       l,
+		r:       r,
+		reg:     reg,
+		st:      st,
+		lim:     lim,
+		enabled: enabled,
+		m:       &APIMetrics{}}
 	a.initMetrics()
 	return a
 }
@@ -172,22 +177,37 @@ func (a *API) registerValidator(w http.ResponseWriter, r *http.Request) {
 }
 
 func (a *API) getHeader(w http.ResponseWriter, r *http.Request) {
-	w.Header().Set("Content-Type", "application/json")
+	if !a.enabled.GetHeader {
+		if ErrorsOnDisable {
+			w.WriteHeader(http.StatusForbidden)
+			a.m.ApiReqCounter.WithLabelValues("getHeader", "403", "forbiden").Inc()
+		} else {
+			a.m.ApiReqCounter.WithLabelValues("getHeader", "499", "disabled").Inc()
+			writeError(w, http.StatusBadRequest, errors.New("no builder bid"))
+		}
+		return
+	}
 
+	w.Header().Set("Content-Type", "application/json")
 	timer := prometheus.NewTimer(a.m.ApiReqTiming.WithLabelValues("getHeader"))
 	defer timer.ObserveDuration()
 
+	uc := structs.UserContent{IP: r.Header.Get("X-Forwarded-For")}
+	var l = a.l.With(log.F{
+		"ip":       uc.IP,
+		"endpoint": "getHeader",
+	})
+
 	req := ParseHeaderRequest(r)
 	m := structs.NewMetricGroup(4)
-	response, err := a.r.GetHeader(r.Context(), m, req)
+	response, err := a.r.GetHeader(r.Context(), m, uc, req)
 	if err != nil {
 		m.ObserveWithError(a.m.RelayTiming, unwrapError(err, "get header unknown"))
 		a.m.ApiReqCounter.WithLabelValues("getHeader", "400", "get header").Inc()
 		slot, _ := req.Slot()
 		proposer, _ := req.Pubkey()
-		a.l.With(log.F{
+		l.With(log.F{
 			"code":     400,
-			"endpoint": "getHeader",
 			"payload":  req,
 			"slot":     slot,
 			"proposer": proposer,
@@ -199,7 +219,7 @@ func (a *API) getHeader(w http.ResponseWriter, r *http.Request) {
 	m.Observe(a.m.RelayTiming)
 
 	if err = json.NewEncoder(w).Encode(response); err != nil {
-		a.l.WithError(err).WithField("path", r.URL.Path).Debug("failed to write response")
+		l.WithError(err).WithField("path", r.URL.Path).Debug("failed to write response")
 		a.m.ApiReqCounter.WithLabelValues("getHeader", "500", "response encode").Inc()
 		// we don't write response as encoder already crashed
 		return
@@ -209,7 +229,19 @@ func (a *API) getHeader(w http.ResponseWriter, r *http.Request) {
 }
 
 func (a *API) getPayload(w http.ResponseWriter, r *http.Request) {
+	if !a.enabled.GetPayload {
+		w.WriteHeader(http.StatusForbidden)
+		a.m.ApiReqCounter.WithLabelValues("getHeader", "403", "get header").Inc()
+		return
+	}
+
 	w.Header().Set("Content-Type", "application/json")
+
+	uc := structs.UserContent{IP: r.Header.Get("X-Forwarded-For")}
+	var l = a.l.With(log.F{
+		"ip":       uc.IP,
+		"endpoint": "getPayload",
+	})
 
 	timer := prometheus.NewTimer(a.m.ApiReqTiming.WithLabelValues("getPayload"))
 	defer timer.ObserveDuration()
@@ -236,12 +268,11 @@ func (a *API) getPayload(w http.ResponseWriter, r *http.Request) {
 		req = &creq
 		if !creq.Validate() {
 			a.m.ApiReqCounter.WithLabelValues("getPayload", "400", "payload validation").Inc()
-			a.l.With(log.F{
+			l.With(log.F{
 				"code":      400,
-				"endpoint":  "getPayload",
 				"slot":      creq.Slot(),
 				"blockHash": creq.BlockHash(),
-			}).Debug("invalid payload")
+			}).With(req).Debug("invalid payload")
 			writeError(w, http.StatusBadRequest, errors.New("invalid payload"))
 			return
 		}
@@ -256,12 +287,11 @@ func (a *API) getPayload(w http.ResponseWriter, r *http.Request) {
 		req = &breq
 		if !breq.Validate() {
 			a.m.ApiReqCounter.WithLabelValues("getPayload", "400", "payload validation").Inc()
-			a.l.With(log.F{
+			l.With(log.F{
 				"code":      400,
-				"endpoint":  "getPayload",
 				"slot":      breq.Slot(),
 				"blockHash": breq.BlockHash(),
-			}).Debug("invalid payload")
+			}).With(req).Debug("invalid payload")
 			writeError(w, http.StatusBadRequest, errors.New("invalid payload"))
 			return
 		}
@@ -271,16 +301,15 @@ func (a *API) getPayload(w http.ResponseWriter, r *http.Request) {
 	}
 
 	m := structs.NewMetricGroup(4)
-	payload, err := a.r.GetPayload(r.Context(), m, req)
+	payload, err := a.r.GetPayload(r.Context(), m, uc, req)
 	if err != nil {
 		m.ObserveWithError(a.m.RelayTiming, unwrapError(err, "get payload unknown"))
 		a.m.ApiReqCounter.WithLabelValues("getPayload", "400", "get payload").Inc()
-		a.l.With(log.F{
+		l.With(log.F{
 			"code":      400,
-			"endpoint":  "getPayload",
 			"slot":      req.Slot(),
 			"blockHash": req.BlockHash(),
-		}).WithError(err).Debug("failed getPayload")
+		}).With(req).WithError(err).Debug("failed getPayload")
 		writeError(w, http.StatusBadRequest, err)
 		return
 	}
@@ -288,7 +317,7 @@ func (a *API) getPayload(w http.ResponseWriter, r *http.Request) {
 	m.Observe(a.m.RelayTiming)
 
 	if err := json.NewEncoder(w).Encode(payload); err != nil {
-		a.l.WithError(err).WithField("path", r.URL.Path).Debug("failed to write response")
+		l.WithError(err).WithField("path", r.URL.Path).Debug("failed to write response")
 		a.m.ApiReqCounter.WithLabelValues("getPayload", "500", "encode response").Inc()
 		// we don't write response as encoder already crashed
 		return
@@ -299,8 +328,20 @@ func (a *API) getPayload(w http.ResponseWriter, r *http.Request) {
 
 // builder related handlers
 func (a *API) submitBlock(w http.ResponseWriter, r *http.Request) {
+	if !a.enabled.SubmitBlock {
+		if ErrorsOnDisable {
+			w.WriteHeader(http.StatusForbidden)
+			a.m.ApiReqCounter.WithLabelValues("submitBlock", "403", "forbidden").Inc()
+		} else {
+			a.m.ApiReqCounter.WithLabelValues("submitBlock", "499", "disabled").Inc()
+		}
+		return
+	}
+
 	w.Header().Set("Content-Type", "application/json")
-	var l = a.l
+	uc := structs.UserContent{IP: r.Header.Get("X-Forwarded-For")}
+	var l = a.l.With(log.F{"ip": uc.IP})
+
 	timer := prometheus.NewTimer(a.m.ApiReqTiming.WithLabelValues("submitBlock"))
 	defer timer.ObserveDuration()
 	var req structs.SubmitBlockRequest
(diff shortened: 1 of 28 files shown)


relay/relay.go
```diff
@@ -383,6 +392,14 @@ func (rs *Relay) GetPayload(ctx context.Context, m *structs.MetricGroup, payload
 		"processingTimeMs": time.Since(tStart).Milliseconds(),
 	})
 
+	go func() {
```
lukanus, line 395: Not a goroutine here, just write to the channel (can be hidden inside api). Also what we would like to save is a raw payload that someone sent to us. So that we can capture code failures

relay/relay.go
```diff
@@ -110,6 +114,8 @@ type RelayConfig struct {
 
 	PublishBlock bool
 
+	ExportData bool
```
lukanus, line 117: You can get that from DataExporter being nil.

relay/relay.go
```diff
@@ -133,6 +139,8 @@ type Relay struct {
 	beacon      Beacon
 	beaconState State
 
+	exp DataExporter
```
lukanus, line 142: To be fair I don't think that export is the right word for that. It's not an export - it's just another kind of storage. Persistor maybe? Or there is definitely better wording for that

data/structs.go
```diff
@@ -0,0 +1,51 @@
...
+
+type exportRequest struct {
+	dt     DataType
+	data   any
+	caller string
+	err    chan error
+}
+
+type exportFiles struct {
+	BlockBidAndTrace *os.File
```
lukanus, line 28: Nothing in this file depends on the type, so why bother?

data/exporter.go
```diff
@@ -0,0 +1,118 @@
...
+		// create files
+		// // BlockBidAndTrace
+		filename := fmt.Sprintf("%s/blockBidAndTrace/output_%d.json", s.datadir, i)
+		file, err := os.Create(filename)
+		if err != nil {
+			return fmt.Errorf("failed to create file: %w", err)
+		}
+		files.BlockBidAndTrace = file
+
+		go func(ctx context.Context, i int, files exportFiles) {
```
lukanus, line 45: definitely separate non anonymous function

data/exporter.go
```diff
@@ -0,0 +1,118 @@
...
+	for i := 0; i < numWorkers; i++ {
+		files := exportFiles{}
+		// create files
+		// // BlockBidAndTrace
+		filename := fmt.Sprintf("%s/blockBidAndTrace/output_%d.json", s.datadir, i)
+		file, err := os.Create(filename)
+		if err != nil {
+			return fmt.Errorf("failed to create file: %w", err)
+		}
+		files.BlockBidAndTrace = file
```
lukanus, line 43: This has to be called inside worker that would take full temporal ownership of the file handler - as I explained. This way you don't deal either with reading and writing from/to the same file by external process

cmd/dreamboat/main.go
```diff
@@ -438,6 +459,22 @@ func run() cli.ActionFunc {
 			return err
 		}
 
+		var exporter relay.DataExporter
+		if c.Bool("data-export") {
```
lukanus, line 463: A presence of non-zero data-export-workers (that should be 0 by default) would be more than enough to enable that

data/exporter.go
```diff
@@ -0,0 +1,118 @@
...
+	request := exportRequest{dt: BlockBidAndTraceData, data: bbt, caller: caller, err: make(chan error, 1)}
+
+	// submit request
+	select {
+	case s.requests <- request:
+	case <-ctx.Done():
+		return ctx.Err()
+	}
+
+	// wait for response
```
lukanus, line 111: We don't wait for response that's fully async, if there is any problem with writing - we deal with this (including logs) inside workers

data/exporter.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+}
+
+func (s ExportService) SubmitBlockBidAndTrace(ctx context.Context, bbt structs.BlockBidAndTrace, caller string) error {
+	request := exportRequest{dt: BlockBidAndTraceData, data: bbt, caller: caller, err: make(chan error, 1)}
+
+	// submit request
+	select {
+	case s.requests <- request:
+	case <-ctx.Done():
```
lukanus, line 107: Probably still wanna store that on shutting down.

data/exporter.go
```diff
@@ -0,0 +1,118 @@
...
+			if err != nil {
+				logger.WithError(err).Error("failed to export request")
+			}
+
+			data := dataWithCaller{Data: req.data, Caller: req.caller}
+			select {
+			case req.err <- enc.Encode(data): // does not block because it is buffered (1) channel, but better safe than sorry
+			case <-ctx.Done():
+				logger.WithError(ctx.Err()).Error("failed to export request")
+			}
```
lukanus, line 94: This suppose to be just a writer. 
It has to do: 
1. Check the request for type (payload, block submission etc)
2. Check request for primary sequence number (slot, block etc) - assigned externally.
3. Get or open filehandler by {type}-{slot}-{worker_id} ->  /data/blah/payload-123456-4  the handler should preferably append only. 
4. Append to that file using piped writers. 
      - As it expects a lot of data I'd recommend gzip + base64 to the payload.  
      - We need to have timestamp of an event so to the byte series, I would prepend nanosecond timestamp with separator.
      - There is a chance we may need additional identifier - that's optional nice to have primary key that we can use to filter out data 
      - the line written can be for example `1680596297123456;0x12345678902345678;{payload}/n`
      - Close filehandler after given time of nothing writing to it.

data/exporter.go
aratz-lasa: If it's dealed by the worker, then we cannot fail on initialization. But we need to propagate the error into `main`, once the workers start running

relay/relay.go
aratz-lasa: If the data-exporter buffer is full, or the writers are working on it, then the `GetPayload` response latency increases. That should never happen. Exporting data should not affect to the time it takes to respond `GetPayload`

relay/relay.go
aratz-lasa: I understand, and you are right, but what is the purpose of this? The trace is for the external world; this data is internally for analyzing and debugging. Any name based on that?

data/structs.go
aratz-lasa: I don't understand the suggestion

data/exporter.go
lukanus: If you like to test if given path is writeable - it's fine to do so on the startup (like touch).
But in this case - we would definitely like to have a size limitted (rotated by relay) files. As this ones would be problematic to deal with after submission

relay/relay.go
lukanus: Datastore? ;) I would say DataWriter would be better

data/exporter.go
aratz-lasa: Even if we don't do it fully async, the writer buffers may cause high-latency. So I would recommend doing this sync and calling it from a goroutine.

relay/relay.go
lukanus: Below we are already storing the payload in goroutine after publishing it to maybe reuse that one, or move data api to the common one.

relay/relay.go
```diff
@@ -99,6 +99,10 @@ type Beacon interface {
 	PublishBlock(ctx context.Context, block structs.SignedBeaconBlock) error
 }
 
+type DataExporter interface {
+	SubmitBlockBidAndTrace(ctx context.Context, bbt structs.BlockBidAndTrace, caller string) error
```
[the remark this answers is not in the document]
aratz-lasa: If we call it `Store`, we cannot extend it to other types. We will need to rename it

data/structs.go
lukanus: Make this universal don't call it "BlockBidAndTrace" but rather "Handler".  ro something

relay/relay.go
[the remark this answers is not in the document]
aratz-lasa: Unless we use `any`, but then the typing is useless in go

data/exporter.go
lukanus: Then let's just use logs and stdout  :)

data/exporter.go
aratz-lasa: > {type}-{slot}-{worker_id}
Why this, inseatd of {type}-{worker_id}?

> There is a chance we may need additional identifier
What are you thinking exactly?

lukanus: In geyPayload it would be most likely block hash or proposer pubkey.  In submit  either block hash or builder  (?)  pubkey

relay/relay.go
```diff
@@ -383,6 +389,16 @@ func (rs *Relay) GetPayload(ctx context.Context, m *structs.MetricGroup, payload
 		"processingTimeMs": time.Since(tStart).Milliseconds(),
 	})
 
+	if rs.exp != nil {
+		go func() {
```
lukanus, line 393: Can we create deferred goroutine that would run both SubmitGetPayloadRequest and PutDelivered based on the flags? This way we'll have half of them running, and we don't need this to be saved immeditely :)

data/exporter.go
```diff
@@ -0,0 +1,199 @@
...
+	logger := s.logger.WithField("id", id)
+
+	logger.Info("started")
+	defer logger.Info("stopped")
+
+	worker := newWorker(id, datadir, logger)
+	ticker := time.NewTicker(filesPruneTick)
+	for {
+		select {
+		case req := <-s.requests:
```
lukanus, line 49: Definitely a great candidate to some metrics, don't you think? :)

data/exporter.go
```diff
@@ -0,0 +1,199 @@
...
+	// get
+	filename := fmt.Sprintf("%s/%s/output_%d_%d.json", w.datadir, toString(req.dt), req.slot, w.id)
+	fileWithTs, ok := w.files[filename]
+	if ok {
+		fileWithTs.ts = time.Now()
+		return fileWithTs.File, nil
+	}
+
+	// create
+	file, err := os.Create(filename)
```
lukanus, line 171: Create truncates file if already exists. That would mean it after the restart it would erase all the data we have for such slot/worker pair. You need to check if the file exists and open that or create the new one. Also for opening file you probably like to use os.O_APPEND

data/exporter.go
```diff
@@ -0,0 +1,199 @@
...
+	case err := <-request.err:
+		return err
+	case <-ctx.Done():
+		return ctx.Err()
+	}
+}
+
+func writeToFile(file *os.File, req exportRequest) error {
+	// Encode struct as JSON and write to base64 encoder
+	_, err := file.Write([]byte(strconv.Itoa(req.timestamp.Nanosecond())))
```
lukanus, line 99: To save some time from this thread - not much, but as this one would be rather hard working - you can create this identifier where you create exportRequest structure.  In this case that would be `relay.go`/`submit.go` -  this way here, you can have it prepared just to be written as one operation :)

data/exporter.go
```diff
@@ -0,0 +1,199 @@
...
+			worker.closeIdleFiles()
+		case <-ctx.Done():
+			worker.closeIdleFiles()
+			return
+		}
+	}
+}
+
+func (s *ExportService) SubmitGetPayloadRequest(ctx context.Context, gpr structs.SignedBlindedBeaconBlock) error {
+	request := exportRequest{dt: BlockBidAndTraceData, data: gpr.Raw(), slot: gpr.Slot(), timestamp: time.Now(), id: gpr.BlockHash().String(), err: make(chan error, 1)}
```
lukanus, line 79: I think would be better if timestamp here would be the one then it happened, rather than when it was marshaled to be stored. 
Also you really prefer it like that, and not just `Store(ctx context.Context, payload  exportRequest)` ? 
We if can assign this values externally - there is no need for /data/exporter to know about any types.

relay/relay.go
aratz-lasa: No, we cannot, really, because `SubmitGetPayloadRequest` occurs even if the request is not successful, while `PutDelivered` only when it's successful

lukanus: You can gather flags from both actions so like are we erroring or not. and run that goroutine inside defer, so that you can - inside goroutine - skip PutDelivered if that wasn't

datastore/warehouse/warehouse.go
```diff
@@ -0,0 +1,278 @@
...
+func (s *Warehouse) RunParallel(ctx context.Context, datadir string, numWorkers int) error {
+	for i := 0; i < numWorkers; i++ {
+		go s.Run(ctx, datadir, i)
+	}
+
+	return nil
+}
+
+func (s *Warehouse) Run(ctx context.Context, datadir string, id int) {
+	s.runningWorkers.Add(1)
```
lukanus, line 47: may be a nitpick but call add outside :)

relay/relay.go
```diff
@@ -395,6 +403,25 @@ func (rs *Relay) GetPayload(ctx context.Context, m *structs.MetricGroup, payload
 		"processingTimeMs": time.Since(tStart).Milliseconds(),
 	})
 
+	var (
+		storeRequest = rs.wh != nil
+		storeTrace   = false
+	)
+	defer func() {
```
lukanus, line 410: 👍

relay/relay.go
```diff
@@ -468,30 +483,31 @@ func (rs *Relay) GetPayload(ctx context.Context, m *structs.MetricGroup, payload
...
+	}
 
-func (wg *TimeoutWaitGroup) Add(i int64) {
-	select {
-	case <-wg.done:
+	if err := rs.wh.Store(context.Background(), req); err != nil {
+		logger.WithError(err).Error("failed to export")
 		return
-	default:
+	} else {
```
lukanus, line 497: else not needed

structs/forks/bellatrix/bellatrix.go
```diff
@@ -25,6 +26,14 @@ func (b *SubmitBlockRequest) Validate() bool {
 		b.BellatrixExecutionPayload.EpTimestamp > 0
 }
 
+func (b *SubmitBlockRequest) Raw() []byte {
+	return b.BellatrixRaw
+}
+
+func (b *SubmitBlockRequest) ExecutionPayload() structs.ExecutionPayload {
```
lukanus, line 33: I think you don't need that

datastore/warehouse/warehouse.go
```diff
@@ -0,0 +1,278 @@
...
+func openOrCreateFile(filename string) (*os.File, error) {
+	var file *os.File
+
+	// check if file exists
+	if _, err := os.Stat(filename); os.IsNotExist(err) {
+		// file doesn't exist, create a new one
+		file, err = os.Create(filename)
+		if err != nil {
+			return nil, err
+		}
```
lukanus, line 268: return, remove else

datastore/warehouse/warehouse.go
```diff
@@ -0,0 +1,278 @@
...
+				w.logger.WithError(err).WithField("filename", filename).Error("failed to close file")
+			}
+			delete(w.files, filename)
+		}
+	}
+}
+
+func (w *worker) closeAllFiles() {
+	for filename, file := range w.files {
+		if err := file.Close(); err != nil {
```
lukanus, line 242: I think you can call closeFile(file) here

datastore/warehouse/warehouse.go
```diff
@@ -0,0 +1,278 @@
...
+	}
+
+	w.files[filename] = fileWithTimestamp{File: file, ts: time.Now()}
+	return file, nil
+}
+
+func (w *worker) closeIdleFiles() {
+	for filename, file := range w.files {
+		if time.Since(file.ts) > fileIdleTime {
+			if err := file.Close(); err != nil {
```
lukanus, line 232: ... and closeFile(file) here as well

datastore/warehouse/warehouse.go
```diff
@@ -0,0 +1,278 @@
...
+
+func newWorker(id int, datadir string, logger log.Logger) *worker {
+	return &worker{id: id, datadir: datadir, files: make(map[string]fileWithTimestamp), logger: logger}
+}
+
+func (w *worker) getOrCreateFile(req StoreRequest) (*os.File, error) {
+	// get
+	filename := fmt.Sprintf("%s/%s/output_%d_%d.json", w.datadir, toString(req.DataType), req.Slot, w.id)
+	fileWithTs, ok := w.files[filename]
+	if ok {
```
lukanus, line 214:
```suggestion
	if fileWithTs, ok := w.files[filename]; ok {

```

datastore/warehouse/warehouse.go
```diff
@@ -0,0 +1,278 @@
...
+		if err != nil {
+			logger.WithError(err).With(req).Error("failed to write")
+
+			file.Write([]byte("\n"))
+			w.closeFile(file)
+			logger.WithField("filename", file.Name()).Debug("corrupted file closed")
+		}
+	} else {
+		logger.WithError(err).Error("failed to get/create file: %w", err)
+	}
```
lukanus, line 91:
```suggestion
	file, err := w.getOrCreateFile(req)
	if err != nil {
		logger.WithError(err).Error("failed to get/create file: %w", err)
		 return // or something
	}
	if err = writeToFile(file, req); err != nil {
		logger.WithError(err).With(req).Error("failed to write")

		file.Write([]byte("\n"))
		w.closeFile(file)
		logger.WithField("filename", file.Name()).Debug("corrupted file closed")
	        return // or something
	}
	
```

datastore/warehouse/warehouse.go
```diff
@@ -0,0 +1,278 @@
...
+	select {
+	case req.err <- err:
+		return
+	case <-s.shutdown:
+		logger.WithError(ErrClosed).Error("failed to export request")
+		return
+	case <-ctx.Done():
+		logger.WithError(ctx.Err()).Error("failed to export request")
+		return
+	}
```
lukanus, line 113: wouldn't life be easier if you just return, and do whatever is needed one level above? this one is never called asynchronously

datastore/warehouse/warehouse.go
```diff
@@ -0,0 +1,278 @@
...
+}
+
+func (s *Warehouse) StoreAsync(ctx context.Context, req StoreRequest) error {
+	// check if service is closed
+	select {
+	case <-s.shutdown:
+		return ErrClosed
+	default:
+	}
```
lukanus, line 148: if we offloaded something to get asynchronously stored, we probably don't like to loose it :) I think the strategy should be different. So, wait until everything that was requested was finished, "however long it would take" and then finish. So I would remove this and also 154-155

relay/submit.go
```diff
@@ -97,6 +98,22 @@ func (rs *Relay) SubmitBlock(ctx context.Context, m *structs.MetricGroup, sbr st
 		return err
 	}
 
+	if rs.wh != nil {
+		go func() {
```
lukanus, line 102: :shipit:  ... pssss look what's happening on the 247  :)

structs/forks/bellatrix/bellatrix.go
```diff
@@ -25,6 +26,14 @@ func (b *SubmitBlockRequest) Validate() bool {
 		b.BellatrixExecutionPayload.EpTimestamp > 0
 }
 
+func (b *SubmitBlockRequest) Raw() []byte {
+	return b.BellatrixRaw
```
lukanus, line 30: You got me. I have no idea where you get this from : > I thought you would need to change api.go to have that available. Or extend unmarshal

datastore/warehouse/warehouse.go
```diff
@@ -0,0 +1,278 @@
...
+
+	// wait for response
+	select {
+	case err := <-req.err:
+		return err
+	case <-s.shutdown:
+		return ErrClosed
+	case <-ctx.Done():
+		return ctx.Err()
+	}
```
lukanus, line 139: You already have it in 151

datastore/warehouse/warehouse.go
aratz-lasa: I would not do that in case someone wants to call `Run()` directly, instead of `RunParallel()`

api/api.go
```diff
@@ -293,6 +307,15 @@ func (a *API) submitBlock(w http.ResponseWriter, r *http.Request) {
...
 	var req structs.SubmitBlockRequest
 
+	b, err := ioutil.ReadAll(r.Body)
+	if err != nil {
+		a.m.ApiReqCounter.WithLabelValues("getPayload", "400", "read body").Inc()
+		writeError(w, http.StatusBadRequest, errors.New("unable to read request body"))
+		return
+	}
+	// Reset the request body's read position
+	r.Body = ioutil.NopCloser(bytes.NewBuffer(b))
```
lukanus, line 317: same as above

datastore/warehouse/warehouse.go
```diff
@@ -0,0 +1,256 @@
...
+	"strconv"
+	"time"
+
+	"github.com/blocknative/dreamboat/structs"
+	"github.com/lthibault/log"
+)
+
+var (
+	fileIdleTime   = structs.DurationPerSlot
+	filesPruneTick = fileIdleTime + (fileIdleTime / 2) // x 1.5
```
lukanus, line 21: nitpick.  filesCloseTick or something. As we're not removing anything here

relay/submit.go
```diff
@@ -98,6 +98,25 @@ func (rs *Relay) SubmitBlock(ctx context.Context, m *structs.MetricGroup, uc str
 		return err
 	}
 
+	// TODO
```
lukanus, line 101: ; > ?

structs/forks/bellatrix/bellatrix.go
```diff
@@ -13,6 +13,7 @@ import (
 
 // BuilderSubmitBlockRequest spec: https://flashbots.notion.site/Relay-API-Spec-5fb0819366954962bc02e81cb33840f5#fa719683d4ae4a57bc3bf60e138b0dc6
 type SubmitBlockRequest struct {
+	BellatrixRaw              []byte           `json:"-"`
```
lukanus, line 16: have we checked it's not a problem in SSZ state root calculations? (asking for a friend ;> )

datastore/warehouse/structs.go
```diff
@@ -0,0 +1,53 @@
...
+		return "GetPayloadRequest"
+	case SubmitBlockRequest:
+		return "SubmitBlockRequest"
+	default:
+		return "unknown"
+	}
+}
+
+type StoreRequest struct {
+	DataType  DataType
```
lukanus, line 32: Assuming that we would like to introduce that in more places - do we need this to be defined defined? 
You're not really switching on that, it's only used for logging. `DataType string` would be more than enough. and as we pass A lot already, few more is not tragedy

datastore/warehouse/structs.go
```diff
@@ -0,0 +1,53 @@
...
+	Data      []byte
+	Slot      uint64
+	Id        string
+	Timestamp time.Time
+
+	header []byte
```
lukanus, line 38: Doesn't look used :)

datastore/warehouse/warehouse.go
```diff
@@ -0,0 +1,251 @@
...
+	if _, err := os.Stat(filedir); os.IsNotExist(err) {
+		err = os.MkdirAll(filedir, 0755)
+		if err != nil {
+			return nil, fmt.Errorf("failed to create directory: %w", err)
+		}
+	}
+
+	// check if file exists
+	if _, err := os.Stat(filename); os.IsNotExist(err) {
+		return os.Create(filename)
```
lukanus, line 241:
```
file, err := os.OpenFile(filename, os.O_CREATE|os.O_APPEND|os.O_WRONLY, 0644)
if err != nil {
			return nil, fmt.Errorf("failed creating files: %w", err)
		}
return file
```

relay/submit.go
aratz-lasa: TODO is because we have to enable this in the future.

datastore/warehouse/warehouse.go
```diff
@@ -0,0 +1,251 @@
...
+
+func (s *Warehouse) StoreAsync(ctx context.Context, req StoreRequest) error {
+	// pre-calculate header of the data
+	var header bytes.Buffer
+	header.WriteString("\n")
+	header.WriteString(strconv.Itoa(int(req.Timestamp.UnixNano())))
+	header.WriteString(";")
+	header.WriteString(req.Id)
+	header.WriteString(";")
+	req.header = header.Bytes()
```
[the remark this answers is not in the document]
aratz-lasa: I prefer having write than append; if it's gonna be multiline anyway, it's more readable.

structs/forks/bellatrix/bellatrix.go
aratz-lasa: How do we check this? In the testnet and Goerli the payloads are being delivered. That should be enough evidence

datastore/warehouse/structs.go
aratz-lasa: Very good idea!

datastore/warehouse/structs.go
aratz-lasa: It's used in `writeToFile`

datastore/warehouse/warehouse.go
```diff
@@ -0,0 +1,251 @@
...
+WorkerLoop:
+	for {
+		select {
+		case req := <-s.requests:
+			s.handleRequest(ctx, logger, w, req)
+			continue
+		case <-pruneTicker.C:
+			w.closeIdleFiles()
+			continue
+		case <-ctx.Done():
```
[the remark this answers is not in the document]
aratz-lasa: Well, we discussed this some weeks ago. The context should be canceled after the HTTP server is closed. 

But of course, we can do a final loop to process all remaining request in the channel.

datastore/warehouse/structs.go
lukanus: Can we make it Public / Upercase then so it reads better? :)

aratz-lasa: We could, but I do not recommend it, because it should not be used by the relay component. Header is used for the internal optimization of warehouse.

lukanus: 👍

lukanus (approved):

PR #127, opened 2023-04-04
GetPayload check for rejecting late requests

aratz-lasa:
# What 🕵️‍♀️
Add a check for rejecting GetPayload requests that occur too late


lukanus (approved):
Not sure why is that needed but ok.

relay/relay.go
```diff
@@ -20,6 +20,10 @@ import (
 	"github.com/blocknative/dreamboat/verify"
 )
 
+const (
+	GetPayloadRequestLimit = 10 * time.Second
```
lukanus, line 24: Let's make it configurable and merge this one

PR #130, opened 2023-04-04
Patches for getpayload

aratz-lasa:
# What 🕵️‍♀️
This PR implements extra security patches for GetPayload, which are: random delay [0-500ms] and do not allow duplicate GetPayload requests.

# Why 🔑
This is necessary for improving security on GetPayload


lukanus (approved):
:)

lukanus (approved):

PR #131, opened 2023-04-05
Add logic to manually fetch header if subscription fails

aratz-lasa:
# What 🕵️‍♀️
This PR implements the logic to manually fetch the latest header in case the subscription fails to receive the event

# Why 🔑
This way, we do not always need to rely on the subscription. If it fails, we have a mechanism to fetch it manually, and keep the relay synced

### DIFF
--- a/beacon/client/client.go
+++ b/beacon/client/client.go
@@ -367,7 +367,7 @@ func (h HeadEvent) Loggable() map[string]any {
 
 // Header is the block header from the beacon chain
 type Message struct {
-	Slot          uint64 `json:"slot"`
+	Slot          uint64 `json:"slot,string"`
 	ProposerIndex string `json:"proposer_index"`
 	ParentRoot    string `json:"parent_root"`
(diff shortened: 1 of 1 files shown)


beacon/client/client.go
```diff
@@ -60,13 +60,23 @@ func (b *beaconClient) SubscribeToHeadEvents(ctx context.Context, slotC chan Hea
...
-				break EventSelect
+				logger.Warn("timed out head events subscription, restarting and manually querying latest header")
+
+				go b.manuallyFetchLatestHeader(ctx, logger, slotC)
+
+				// to prevent disconnection due to multiple timeouts occurring very close in time, the subscription loop is canceled and restarted
+				if time.Since(lastTimeout) <= BeaconEventTimeout*2 {
+					cancelLoop()
+					break EventSelect
+				} else {
```
lukanus, line 76: you don't need else here as first condition finishes with break

lukanus (approved):

PR #136, opened 2023-04-06
Improve beacon subscription restart and logs

aratz-lasa:
# What 🕵️‍♀️
This PR improves logic for beacon subscription restart so that it checks whether there are 3 consecutive timeouts. If they are not consecutive and close in time, then it does not restart, it only fetches the latest header manually.

# Why 🔑
This avoids overly restarting subscriptions in networks like Goerli, where there are many missed slots per epoch.

### DIFF
--- a/beacon/client/client.go
+++ b/beacon/client/client.go
@@ -76,7 +76,7 @@ func (b *beaconClient) SubscribeToHeadEvents(ctx context.Context, slotC chan Hea
 				if time.Since(lastTimeout) <= BeaconEventTimeout*2 {
 					timeoutCounter++
 					if timeoutCounter >= BeaconConsecutiveTimeout {
-						logger.WithField("timeoutCounter", timeoutCounter).Debug("restarting subcription")
+						logger.WithField("timeoutCounter", timeoutCounter).Warn("restarting subcription")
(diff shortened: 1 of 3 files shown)


beacon/client/client.go
```diff
@@ -19,8 +19,9 @@ import (
 )
 
 const (
-	BeaconEventTimeout = structs.DurationPerSlot + (structs.DurationPerSlot / 4)
-	BeaconQueryTimeout = 20 * time.Second
+	BeaconEventTimeout       = structs.DurationPerSlot + (structs.DurationPerSlot / 4)
```
lukanus, line 22: Shouldn't we make is configurable, not to have another PR tomorrow ? :) The "4" i mean

lukanus (approved):

beacon/client/client.go
aratz-lasa: The BeaconEventTimeout or the BeaconConsecutiveTimeout?

lukanus: To be fair both look like good candidates, as we know that's different on each one of 3 chains

aratz-lasa: Nice, I will

PR #138, opened 2023-04-07
Payload validation on GetPayload

aratz-lasa:
# What 🕵️‍♀️
Add header validation to the GetPayload request so that the header sent in the request and the one the relay has been the same.

# Why 🔑
This is necessary for new security issue on malicious proposers

### DIFF
--- a/cmd/dreamboat/main.go
+++ b/cmd/dreamboat/main.go
@@ -241,7 +241,7 @@ var flags = []cli.Flag{
 	},
 	&cli.DurationFlag{
 		Name:    "getpayload-response-delay",
-		Usage:   "Maximum delay between block publication and returning request to validator",
+		Usage:   "Delay between block publication and returning request to validator",
 		Value:   1 * time.Second,
 		EnvVars: []string{"GETPAYLOAD_RESPONSE_DELAY"},
 	},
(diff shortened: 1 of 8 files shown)


cmd/dreamboat/main.go
```diff
@@ -240,10 +240,16 @@ var flags = []cli.Flag{
 		EnvVars: []string{"BLOCK_VALIDATION_ENDPOINT_RPC"},
 	},
 	&cli.DurationFlag{
-		Name:    "max-block-publication-delay",
+		Name:    "getpayload-response-delay",
+		Usage:   "Maximum delay between block publication and returning request to validator",
```
bkellerman, line 244: This doesn't appear to be the `Maximum delay` just the `delay`

cmd/dreamboat/main.go
```diff
@@ -240,10 +240,16 @@ var flags = []cli.Flag{
...
+		Value:   1 * time.Second,
+		EnvVars: []string{"GETPAYLOAD_RESPONSE_DELAY"},
+	},
+	&cli.DurationFlag{
+		Name:    "getpayload-publish-random-delay",
 		Usage:   "Maximum delay between block publication and returning request to validator",
```
bkellerman, line 250: This `Usage` is a dupe of above. Needs to be updated.

lukanus (approved):

PR #139, opened 2023-04-10
Add check for nil pointer on headerHash

aratz-lasa:


lukanus (approved):
If we're  checking for equality and not errors in the other one - all good

PR #142, opened 2023-04-13
Payload sub

aratz-lasa:
# What 🕵️‍♀️
This PR implements subscriptions to beacon for getting payload attributes, instead of waiting for head events and manually fetching them

# Why 🔑
Improves performance, because we do not pull the info, instead it is pushed to us

### DIFF
--- a/beacon/manager.go
+++ b/beacon/manager.go
@@ -247,9 +247,9 @@ func (s *Manager) RunPayloadAttributesSubscription(ctx context.Context, state St
 		logger = logger.WithField("slot", payloadAttributes.Data.ProposalSlot)
 
 		headSlot := state.HeadSlot()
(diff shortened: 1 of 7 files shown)


beacon/manager.go
```diff
@@ -264,23 +338,26 @@ func (s *Manager) processNewSlot(ctx context.Context, state State, client Beacon
 		}(received)
 	}
 
-	// query expected withdrawals root
-	go s.updatedExpectedWithdrawals(headSlot, state, client)
+	if !s.Config.RunPayloadAttributesSubscription {
```
lukanus, line 341: I wonder shouldn't we allow that regardless of payload atributes, like what if payload attributes fails and the other one wont. can we hook it up somehow to our manual requesters?

PR #148, opened 2023-05-02
Add flag for beacons that are only used for publishing blocks

aratz-lasa:
# What 🕵️‍♀️
Adds a new flag to the beacon to separate beacons used for querying the state (e.g. randao), and the beacon nodes used for publishing blocks to the network.

# Why 🔑
Lighthouse and Prysm provide incompatible APIs regarding the state, but they do have the same API for publishing blocks. So we are gonna have separate flags for publishing blocks (which contain Lighthouse and Prysm nodes), and for querying the state (which contains only Prysm nodes)


lukanus (approved):
Check if - in the loggable -  is possible to have nil pointer. Otherwise it's fine

PR #157, opened 2023-05-12
Make stream structs lightweight

aratz-lasa:
# What 🕵️‍♀️
Makes stream lighter, reducing the amount of data is streamed

# Why 🔑
To reduce pressure on the network and pubsub system

### DIFF
--- a/cmd/dreamboat/main.go
+++ b/cmd/dreamboat/main.go
@@ -626,6 +626,7 @@ func run() cli.ActionFunc {
 		if err != nil {
 			return fmt.Errorf("fail to initialize stream cache: %w", err)
 		}
+
 		r := relay.NewRelay(logger, relay.RelayConfig{
 			BuilderSigningDomain:       domainBuilder,
 			GetPayloadResponseDelay:    c.Duration("getpayload-response-delay"),
--- a/cmd/dreamboat/main.go
+++ b/cmd/dreamboat/main.go
@@ -645,6 +645,8 @@ func run() cli.ActionFunc {
 		}, beaconPubCli, validatorCache, valDS, verificator, state, payloadCache, ds, daDS, auctioneer, simFallb, relayWh, streamer)
 		r.AttachMetrics(m)
 
+		r.RunSubscribersParallel(c.Context, c.Uint("relay-distribution-stream-workers"))
+
 		ee := &api.EnabledEndpoints{
 			GetHeader:   true,
 			GetPayload:  true,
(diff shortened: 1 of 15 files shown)


relay/relay.go
```diff
@@ -255,46 +275,64 @@ func (rs *Relay) GetHeader(ctx context.Context, m *structs.MetricGroup, uc struc
...
-	if zero := types.IntToU256(0); header.Trace.Value.Cmp(&zero) == 0 {
+	value := maxProfit.BuilderBid.Value()
+	if zero := types.IntToU256(0); value.Cmp(&zero) == 0 {
 		rs.m.MissHeaderCount.WithLabelValues("zeroBid").Add(1)
 		return nil, ErrZeroBid
 	}
 
 	fork := rs.beaconState.ForkVersion(slot)
+	go func() {
+		if !rs.dc.Contains(key) {
```
lukanus, line 308: That would be Get, not contains, you like to update this get to more recent one

relay/relay.go
```diff
@@ -255,46 +275,64 @@ func (rs *Relay) GetHeader(ctx context.Context, m *structs.MetricGroup, uc struc
...
 
-	if zero := types.IntToU256(0); header.Trace.Value.Cmp(&zero) == 0 {
+	value := maxProfit.BuilderBid.Value()
+	if zero := types.IntToU256(0); value.Cmp(&zero) == 0 {
 		rs.m.MissHeaderCount.WithLabelValues("zeroBid").Add(1)
 		return nil, ErrZeroBid
 	}
 
 	fork := rs.beaconState.ForkVersion(slot)
+	go func() {
```
lukanus, line 307: I'm not sure can we do that in goroutine. as it has to be stored (and probably returned) prior to  returning that to the user. Otherwise there can be a operation where we would not have data stored on getpayload

lukanus: + if we do that we need not to fail on "return"  from GetPayload. Also I would call PublishBlockCache as soon as possible.

stream/structs.go
```diff
@@ -1,141 +1,55 @@
...
+	Source string
 }
 
-type CapellaStreamBlock struct {
-	Block        capella.BlockBidAndTrace
-	IsBlockCache bool
-	StreamSource string
+type JsonItem struct {
+	StreamMeta Metadata
+	StreamData []byte
```
lukanus, line 15: json.RawMessage

cmd/dreamboat/main.go
```diff
@@ -211,12 +211,6 @@ var flags = []cli.Flag{
 		Value:   "",
 		EnvVars: []string{"RELAY_DATAAPI_DATABASE_URL"},
 	},
-	&cli.BoolFlag{ // TODO: Remove
```
lukanus, line 214: Don't remove it just yet unless it's already removed from deployment because it would throw an error about unknown parameter

lukanus (approved):

PR #162, opened 2023-05-17
Block submissions using gzip and/or ssz

aratz-lasa:
## What 🕵️‍♀️
This pull request introduces changes to enhance the codebase and improve the handling of SSZ (Simple Serialize) data and gzip compression. These changes are motivated by the discussions and requirements outlined in [PR #388](https://github.com/flashbots/mev-boost-relay/pull/388/).

## Why 🔑
The need for these changes arises from the requirement to support different content types and improve the overall efficiency and flexibility of the system.

## How ⚙️
The following components were modified or added to achieve the desired enhancements:

1. **SSZ Decoding**:
   - Added support for decoding SSZ-encoded data in the `submitBlock` endpoint.
   - If the request has a content type of `application/octet-stream`, the payload is unmarshaled using the SSZ format.
   - In case of a decoding error, a fallback to JSON decoding is implemented.
   - The SSZ decoding logic is implemented in the `capella.SubmitBlockRequest` struct's `UnmarshalSSZ` method.

2. **Gzip Compression**:
   - Enabled support for gzip compression in the request body of the `submitBlock` endpoint.
   - If the request includes the `Content-Encoding` header with the value "gzip," the request body is read using a gzip reader.
   - The maximum limit for the request body size is increased to 10 MB.
   - The gzip decompression is performed using the `compress/gzip` package.

These modifications enable the system to handle SSZ-encoded payloads and gzip-compressed request bodies, providing better interoperability and efficiency in data transmission.


api/api.go
```diff
@@ -356,11 +374,27 @@ func (a *API) submitBlock(w http.ResponseWriter, r *http.Request) {
 	switch a.st.ForkVersion(a.st.HeadSlot()) {
 	case structs.ForkCapella:
 		var creq capella.SubmitBlockRequest
-		if err := json.NewDecoder(bytes.NewReader(b)).Decode(&creq); err != nil {
-			a.m.ApiReqCounter.WithLabelValues("submitBlock", "400", "payload decode").Inc()
-			writeError(w, http.StatusBadRequest, errors.New("invalid submitblock request capella decode"))
-			return
+		if r.Header.Get("Content-Type") == "application/octet-stream" {
```
lukanus, line 377: There is a chance we may like to do lowercase on that

api/api.go
```diff
@@ -356,11 +374,27 @@ func (a *API) submitBlock(w http.ResponseWriter, r *http.Request) {
...
+			l = l.WithField("requestContentType", "ssz")
+			if err := creq.UnmarshalSSZ(b); err != nil {
+				// Fallback to JSON.
```
lukanus, line 380: Content type is octet-stream. I don't think we should fallback to anything here.

api/api.go
aratz-lasa: Look at the Go code for `http.Post`

Screenshot 2023-05-18 at 11 33 40

lukanus: Yes, I see contentType is string ;> and ? ;)

aratz-lasa: Capital too, not lower case

api/api.go
```diff
@@ -340,13 +342,30 @@ func (a *API) submitBlock(w http.ResponseWriter, r *http.Request) {
 
 	w.Header().Set("Content-Type", "application/json")
 	uc := structs.UserContent{IP: r.Header.Get("X-Forwarded-For")}
-	var l = a.l.With(log.F{"ip": uc.IP})
+	
+	isGzip := strings.ToLower(r.Header.Get("Content-Encoding")) == "gzip"
+	var l = a.l.With(log.F{"ip": uc.IP, "gzip": isGzip})
```
lukanus, line 347: Wouldn't it be better to have "encode" or "compression", and "gzip"/"none" as a type? :)

lukanus (approved):
Good job ! :)

api/api.go
aratz-lasa: a `var` for every encode type you mean?

lukanus: To have information about encoding in logs rather than just "if it's gzip"

lukanus (approved):

PR #184, opened 2023-08-01
Feat/refactor streaming ww prep

lthibault:
This PR refactors `stream.Client` in preparation for Wetware-backed pubsub.  There are no functional changes.

- `stream.Client` is no longer responsible for (un)marshaling the message envelope that is sent to Redis.  This is now the responsibility of the `stream.PubSub` implementation.  A new `stream/transport` package has been created to provide a common envelope implementation called `Message`, and to handle its (de)serialization.
- `stream.Client` is no longer responsible for filtering messages that originate from the local node.  This is now the responsibility of the `stream.PubSub` implementation.
- `stream.PubSub`'s `Subscribe()` method now returns a `transport.Subscriber` interface.  This decoupling makes it easier to perform the aforementioned filtering.  In the near future, it will also facilitate the writing of a Wetware based implementation.
- Instantiation and lifecycle management of `stream.Client` is simplified.  In particular, redundant queues are elided and setup is performed through lazy initialization via `sync.Once`.

@lukanus Apologies for the big commit (141e042e48db9c07a10946319874a88ddd107be6).  Should be tighter from here on out.

### DIFF
--- a/cmd/dreamboat/main.go
+++ b/cmd/dreamboat/main.go
@@ -4,11 +4,11 @@ import (
 	"context"
 	"errors"
 	"flag"
-	"fmt"
 	"math/rand"
 	"net/http"
 	"os"
 	"os/signal"
+	"path"
 	"syscall"
 
 	"time"
@@ -27,7 +27,6 @@ import (
 	"github.com/blocknative/dreamboat/sim"
 	"github.com/blocknative/dreamboat/sim/client/fallback"
 	"github.com/blocknative/dreamboat/stream"
-	"github.com/google/uuid"
 	badger "github.com/ipfs/go-ds-badger2"
 
 	fileS "github.com/blocknative/dreamboat/cmd/dreamboat/config/source/file"
@@ -123,7 +122,7 @@ func main() {
 	var (
 		storage  datastore.TTLStorage
 		badgerDs *badger.Datastore
-		streamer relay.Streamer
+		streamer *stream.Client
 		err      error
 	)
 
@@ -143,24 +142,25 @@ func main() {
 		readClient := redis.NewClient(&redis.Options{
 			Addr: cfg.Payload.Redis.Read.Address,
 		})
+		defer readClient.Close()
 		m.RegisterRedis("redis", "readreplica", readClient)
 
 		writeClient := redis.NewClient(&redis.Options{
 			Addr: cfg.Payload.Redis.Write.Address,
 		})
+		defer writeClient.Close()
 		m.RegisterRedis("redis", "master", writeClient)
 		storage = &dsRedis.RedisDatastore{Read: readClient, Write: writeClient}
 
 		redisClient := redis.NewClient(&redis.Options{
 			Addr: cfg.Distributed.Redis.Address,
 		})
-
+		defer redisClient.Close()
 		m.RegisterRedis("redis", "stream", redisClient)
-		streamer, err = initStreamer(ctx, cfg.Distributed, redisClient, logger, m, state)
-		if err != nil {
-			logger.WithError(err).Error("fail to create streamer")
-			return
-		}
+
+		streamer = newStreamClient(cfg.Distributed, redisClient, logger, m, state)
+		defer streamer.Close()
+
 	} else {
 		storage = badgerDs
 	}
@@ -519,34 +519,52 @@ func ComputeDomain(domainType types.DomainType, forkVersionHex string, genesisVa
 	return types.ComputeDomain(domainType, forkVersion, genesisValidatorsRoot), nil
 }
 
-func initStreamer(ctx context.Context, cfg *config.DistributedConfig, redisClient *redis.Client, l log.Logger, m *metrics.Metrics, st stream.State) (relay.Streamer, error) {
-	timeStreamStart := time.Now()
+func newStreamClient(cfg *config.DistributedConfig, redisClient *redis.Client, l log.Logger, m *metrics.Metrics, st stream.State) *stream.Client {
+	//// TODO:  move this to where the stream subscribers are actually started
+	// timeStreamStart := time.Now()
+	// defer func() {
+	// 	l.WithField("relay-service", "stream-subscriber").
+	// 		WithField("startTimeMs", time.Since(timeStreamStart).Milliseconds()).
+	// 		Info("initialized")
+	// }()
 
-	pubsub := &redisStream.Pubsub{Redis: redisClient, Logger: l}
-
-	id := cfg.InstanceID
-	if id == "" {
-		id = uuid.NewString()
+	return &stream.Client{
+		State: st,
+		Logger: l.
+			WithField("subService", "stream").
+			WithField("type", "redis"),
+		Bids: &redisStream.Topic{
+			Redis:     redisClient,
+			Logger:    l.WithField("topic", stream.BidTopic),
+			Name:      path.Join(cfg.Redis.Topic, stream.BidTopic),
+			LocalNode: cfg.LocalNode(),
+		},
+		Cache: &redisStream.Topic{
+			Redis:     redisClient,
+			Logger:    l.WithField("topic", stream.CacheTopic),
+			Name:      path.Join(cfg.Redis.Topic, stream.CacheTopic),
+			LocalNode: cfg.LocalNode(),
+		},
+		QueueSize:  cfg.StreamQueueSize,
+		NumWorkers: cfg.WorkerNumber,
+		Metrics:    m,
 	}
 
-	streamConfig := stream.StreamConfig{
-		Logger:          l,
-		ID:              id,
-		PubsubTopic:     cfg.Redis.Topic,
-		StreamQueueSize: cfg.StreamQueueSize,
-	}
+	// // ...
 
-	redisStreamer := stream.NewClient(pubsub, st, streamConfig)
-	redisStreamer.AttachMetrics(m)
+	// streamConfig := stream.StreamConfig{
+	// 	Logger:          l,
+	// 	ID:              id,
+	// 	PubsubTopic:     cfg.Redis.Topic,
+	// 	StreamQueueSize: cfg.StreamQueueSize,
+	// }
 
-	if err := redisStreamer.RunSubscriberParallel(ctx, uint(cfg.WorkerNumber)); err != nil {
-		return nil, fmt.Errorf("fail to start stream subscriber: %w", err)
-	}
+	// redisStreamer := stream.NewClient(pubsub, st, streamConfig)
+	// redisStreamer.AttachMetrics(m)
 
-	l.With(log.F{
-		"relay-service": "stream-subscriber",
-		"startTimeMs":   time.Since(timeStreamStart).Milliseconds(),
-	}).Info("initialized")
+	// if err := redisStreamer.RunSubscriberParallel(ctx, uint(cfg.WorkerNumber)); err != nil {
+	// 	return nil, fmt.Errorf("fail to start stream subscriber: %w", err)
+	// }
 
-	return redisStreamer, nil
+	// return redisStreamer, nil
 }
--- a/stream/metrics.go
+++ b/stream/metrics.go
@@ -1,14 +1,13 @@
 package stream
 
 import (
-	"github.com/blocknative/dreamboat/metrics"
 	"github.com/prometheus/client_golang/prometheus"
 )
 
-type StreamMetrics struct {
-	RecvCounter *prometheus.CounterVec
-	Timing      *prometheus.HistogramVec
-	PublishSize *prometheus.HistogramVec
+type streamMetrics struct {
+	RecvCounter    *prometheus.CounterVec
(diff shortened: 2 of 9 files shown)


lthibault:
@lukanus `TestSource_Load/happy` is currently failing, though I'm not sure why. Any ideas?

lthibault:
Note also:  this PR temporarily disables a handful of metrics, as it was awkward to thread them into the newly factored Redis pubsub implementation.  I'll re-enable them in a future PR.

lthibault:
> @lukanus `TestSource_Load/happy` is currently failing, though I'm not sure why. Any ideas?

Ok, turns out a change I made to the config struct invalidates the deep equality check in the test.  This has been fixed by testing for equivalence (rather than pointer identity, as before).

stream/stream.go
```diff
@@ -239,57 +240,59 @@ func (s *Client) PublishBuilderBid(ctx context.Context, bid structs.BuilderBidEx
...
 
 func (s *Client) PublishBlockCache(ctx context.Context, block structs.BlockAndTraceExtended) error {
+	s.init()
+
 	timer0 := prometheus.NewTimer(s.m.Timing.WithLabelValues("publishCacheBlock", "all"))
 
 	timer1 := prometheus.NewTimer(s.m.Timing.WithLabelValues("publishCacheBlock", "encode"))
 	forkEncoding := toBlockCacheFormat(s.st.ForkVersion(structs.Slot(block.Slot())))
-	b, err := s.encode(block, forkEncoding)
+	msg, err := s.encode(block, forkEncoding)
```
lukanus, line 271: Shouldn't this be a a transport function? If you distinct between `Cache` and `Bids` wouldn't that be optimal to pass entire object to publish and let transport internally decide how to encode/carry that?

stream/stream.go
```diff
@@ -4,229 +4,230 @@ package stream
...
-		st:     st,
-
-		builderBidIn:     make(chan []byte, cfg.StreamQueueSize),
-		builderBidOut:    make(chan structs.BuilderBidExtended, cfg.StreamQueueSize),
-		cacheIn:          make(chan []byte, cfg.StreamQueueSize),
-		cacheOut:         make(chan structs.BlockAndTraceExtended, cfg.StreamQueueSize),
-		slotDeliveredIn:  make(chan []byte, cfg.StreamQueueSize),
-		slotDeliveredOut: make(chan uint64, cfg.StreamQueueSize),
+func (c *Client) init() {
+	c.once.Do(func() {
```
lukanus, line 63: I'm not really a fan of once. But regardless - you should be able to instantialize this as much as you like as it doesn't hold global context - I don't see any reason why you would need one here. 
And being - in main.go why would I like to do it on every function call this structure have (BlockCache, BuilderBid,  PublishBuilderBid)?  
New was just meant to safely initialize what ever client may need in future, so that writing any additional function I should not pay attention to some initializer someone may like me to call.

stream/transport/transport.go
```diff
@@ -0,0 +1,71 @@
...
+	if n <= 0 {
+		return Message{}, ErrDecodeVarint
+	}
+
+	b = b[n:]
+
+	switch ForkVersionFormat(varint) {
+	case BellatrixJson, CapellaJson:
+		var msg Message
+		if err := json.Unmarshal(b, &msg); err != nil {
```
lukanus, line 63:
```suggestion 
	switch ForkVersionFormat(varint) {
	case BellatrixJson, CapellaJson:
		var msg Message
		if err := json.Unmarshal(b[n:], &msg); err != nil {
```

stream/metrics.go
```diff
@@ -40,11 +39,9 @@ func (s *Client) initMetrics() {
...
 		Help:      "Number of publications per function",
 	}, []string{"function"})
-}
 
-func (s *Client) AttachMetrics(m *metrics.Metrics) {
-	m.Register(s.m.RecvCounter)
-	m.Register(s.m.Timing)
-	m.Register(s.m.PublishSize)
-	m.Register(s.m.PublishCounter)
+	s.Metrics.Register(s.m.RecvCounter)
```
lukanus, line 43: AttachMetrics it's a pattern used everywhere, if you wanna change it, let's talk about that and do that in separate PR everywhere, so it would not stand out

stream/stream.go
```diff
@@ -4,229 +4,230 @@ package stream
...
+			WithField("method", "runBuilderBidSubscriber").
+			WithField("forkEncoding", forkEncoding).
+			Warn("unkown builder bid forkEncoding")
+		return
+	}
+
+	select {
+	case <-c.ctx.Done():
+	case c.builderBidOut <- bb:
+		c.Logger.With(log.F{
```
lukanus, line 217: There are some logs that I've removed in my pr because of the overwhelming count. That's one of it. We're not using that and just burning cpu on this received and the other "sent" logs are pretty useless. I think we can maybe try commenting that out for now

stream/stream.go
```diff
@@ -4,229 +4,230 @@ package stream
...
+		}
+		bb = &cbb
+	case transport.CapellaSSZ:
+		var cbb capella.BuilderBidExtended
+		if err := cbb.UnmarshalSSZ(msg.Payload); err != nil {
+			c.Logger.WithError(err).
+				WithField("method", "runBuilderBidSubscriber").
+				WithField("forkEncoding", forkEncoding).
+				Warn("failed to decode builder bid")
+			return
```
lukanus, line 203:
```suggestion
func (c *Client) handleBid(msg transport.Message) error {
	var (
		receivedAt = time.Now()
		bb         structs.BuilderBidExtended
	)

	switch forkEncoding := msg.ForkEncoding; forkEncoding {
	case transport.BellatrixJson:
		var bbb bellatrix.BuilderBidExtended
		if err := json.Unmarshal(msg.Payload, &bbb); err != nil {
			return err
		}
		bb = &bbb
	case transport.CapellaJson:
		var cbb capella.BuilderBidExtended
		if err := json.Unmarshal(msg.Payload, &cbb); err != nil {
			return err
		}
		bb = &cbb
	case transport.CapellaSSZ:
		var cbb capella.BuilderBidExtended
		if err := cbb.UnmarshalSSZ(msg.Payload); err != nil {
			return err
```
Nothing important but by returning you can just log it once in subscribe. removing all of this

stream/transport/redis/redis.go
```diff
@@ -2,42 +2,85 @@ package stream
...
+				Warn("failed to get subscription message from redis")
+			time.Sleep(b.Duration())
+			continue
 		}
 
-	}()
+		b.Reset()
 
-	return sub
+		msg, err := transport.Decode([]byte(rawMsg.Payload))
```
lukanus, line 73: In the previous implementation decoding was done in multiple goroutines so that we've been decoding messages in parallel. That's why subscribe was separated from the rest

stream/stream.go
```diff
@@ -4,229 +4,230 @@ package stream
...
-	for i := uint(0); i < num; i++ {
-		go s.RunBlockCacheSubscriber(ctx)
-		go s.RunBuilderBidSubscriber(ctx)
-	}
-
-	go s.RunSlotDeliveredSubscriber(ctx)
-
+func (c *Client) Close() error {
+	c.once.Do(func() {
+		c.ctx, c.cancel = context.WithCancel(context.Background())
```
lukanus, line 81: this may not work if you overwrite c.ctx and close it - for something that was already inherited the previous context downwards

stream/stream.go
lthibault: It could, and likely eventually will.  I'd like to do this in incremental steps, though. For now, it seemed sensible to move the payload envelope (`transport.Message`, formerlly `JsonItem`) to the `transport` package and leave the payload encoding where it is.

stream/transport/redis/redis.go
lthibault: Yeah, I noticed that.  I'm happy to re-introduce this in a follow-up PR, but this one was already getting quite large.  I also suspect that the parallelism isn't needed here, but maybe you've already measured this?

stream/stream.go
lthibault: `sync.Once` will only ever fire once.  It either sets the context here, or in the `init` function.  This is intended to prevent panics due to either (1) nil `c.cancel` in this function or (2) a nil context in a concurrent call to `init`.

stream/metrics.go
lthibault: The idea was to avoid multi-step initialization.  I think this is fine since it's not introducing a diverging API to the end-user; it's all abstracted away and the user sees none of it.

stream/stream.go
lthibault: I'm quite attached to this change because a `New*` function is going to be quite clunky due to the large number of arguments.

> you should be able to instantialize this as much as you like as it doesn't hold global context 

I'm not sure I understand.  The lazy initialization is needed to set up the goroutines and allocate the channels.

>New was just meant to safely initialize what ever client may need in future, so that writing any additional function I should not pay attention to some initializer someone may like me to call.

Functionally, this is exactly equivalent to `New*`.  You just instantiate a struct literal instead of making a function call, but there's no additional step from the point of view of the public API.

stream/stream.go
```diff
@@ -4,229 +4,230 @@ package stream
...
+		}
 
-	s.initMetrics()
+		c.builderBidOut = make(chan structs.BuilderBidExtended, c.QueueSize)
+		c.cacheOut = make(chan structs.BlockAndTraceExtended, c.QueueSize)
 
-	return &s
+		c.ctx, c.cancel = context.WithCancel(context.Background())
+		go c.subscribe(c.Bids, c.handleBid)
+		go c.subscribe(c.Cache, c.handleCache)
```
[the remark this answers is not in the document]
lthibault: I'm not quite sure I follow; do you want to explicitly add subscriptions at the `Client` call site (i.e. in `main.go`)?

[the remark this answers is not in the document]
lukanus: I'm trying to avoid "callback function pattern" here :) 
And it may make sense to just hide `chan builderBidOut` inside another structure - `BidSubscription` in the example above. This way BlockCache method can become a BidSubscription's and can be potentially removed from the Client. 
Maybe `BidSubscriptionHandler` would be a better name?
So from the Client structure perspective, the interface would be just: 
`type SubscribeFor interface {
     Handle(m Message) error
}`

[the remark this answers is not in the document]
lthibault: >And it may make sense to just hide chan builderBidOut inside another structure

This would require each transport implementation to manage the lifecycle of a channel.  In addition to not being very DRY, it also exposes us to leaky abstractions, since each can now have different queuing/concurrency strategies.  So this feels like it would be going in the wrong direction.

I think callbacks are generally fine, provided they're not abused.  It's really no different from providing an interface.

stream/stream.go
lukanus: You call init() in every public function `BlockCache`, `BuilderBid`, `PublishBuilderBid` -  instead of just initializing it once and using from the outside  - not need for knowing/checking anything as you know everything is initialized. And New here exists for the very reason of making all the channels and everything that may possibly effect in nil pointer exception of or deadlock. It also gives you a way not to use it if you, for some reason prefer to bypass that and do it yourself.

stream/transport/redis/redis.go
lukanus: For the parallelism -  It may not be for bids. for payloads however especially that you discard own payloads after the publish it may be signifficant change. As it's up to 300kB per message to decode and we'd know it would grow soon in deneb to +500kB

lthibault: > for payloads however especially that you discard own payloads after the publish it may be signifficant change. 

Just to make sure we're tracking:  we discard payloads that originate from the local node, but the payload doesn't need to be decoded to make this determination.  All that needs to happen is for the envelope (`transport.Message`) to be decoded; the payload is decoded separately.

Therefore, it seems like this would not be very expensive. (Right?)

stream/stream.go
lthibault: e1ee8bef260ce049e95b68f0d131eb65eb10a5f8

cmd/dreamboat/main.go
```diff
@@ -519,34 +518,30 @@ func ComputeDomain(domainType types.DomainType, forkVersionHex string, genesisVa
...
+		Name:      path.Join(cfg.Redis.Topic, stream.CacheTopic),
+		LocalNode: cfg.LocalNode(),
 	}
 
-	redisStreamer := stream.NewClient(pubsub, st, streamConfig)
-	redisStreamer.AttachMetrics(m)
-
-	if err := redisStreamer.RunSubscriberParallel(ctx, uint(cfg.WorkerNumber)); err != nil {
-		return nil, fmt.Errorf("fail to start stream subscriber: %w", err)
+	return &stream.Client{
```
lukanus, line 536: There is a chance you may like to use the NewClient function from below

lthibault: 🤦‍♂️

lukanus (approved):

